PR #2, opened 2025-01-13
Prepare MPSBackendImpl for intra-tdvp checkpoint/autosave

pablolh:
Prepare the code for autosave, which needs to be able to save the state of the simulation for later resuming, even inside tdvp.

For this, the different steps of the tdvp logic are brought inside mps_backend_impl.
Baths are stored inside this same class.
Right baths from a right-left swipe are reused in the following left-right swipe, provided the hamiltonian doesn't change, which occurs in the absence of lindbladian noise.

### DIFF
--- a/emu_base/math/brents_root_finding.py
+++ b/emu_base/math/brents_root_finding.py
@@ -1,4 +1,97 @@
-from typing import Callable
+from typing import Callable, Optional
+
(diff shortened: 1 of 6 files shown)


a-quelle (approved):
LGTM

PR #3, opened 2025-01-13
[Doc] Reworking README, adding LICENSE

Yoric:
Attempting to standardize the README and LICENSE across our new open-source releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,110 +4,64 @@
   # Emu-MPS
 </div>
 
-**EMU-MPS** is a Pulser backend, designed to **EMU**late the dynamics of programmable arrays of neutral atoms, with matrix product states (**MPS**). It allows users to increase the number of qubits and reduce computation time.
+**EMU-MPS** is a backend for the [Pulser low-level Quantum Programming toolkit](https://pulser.readthedocs.io). EMU-MPS lets you transparently run Quantum Algorithms on a simulated device, using GPU acceleration if available. More in depth, EMU-MPS is designed to **EMU**late the dynamics of programmable arrays of neutral atoms, with matrix product states (**MPS**). This design makes it faster and more memory-efficient than previous generations of quantum emulators, which means that you can emulate systems with larger number of qubits, faster.
 
-Join us on [Slack](https://pasqalworkspace.slack.com/archives/C07MUV5K7EU) or by [e-mail](mailto:emulation@pasqal.com) to give us feedback about how you plan to use Emu-MPS or if you require specific feature-upgrades.
-
-
-## Getting started
-
-You can install from source, or download the package from the private pypi registry that pasqal maintains in gitlab.
-For developers, we recommend installing from source, for users we recommend installing from the registry.
+## Installation
 
 **Warning:** installing emu-mps will update pulser-core
 
-We always recommend using a virtual environment.
+### Using `hatch`, `uv` or any pyproject-compatible Python manager
 
-<details>
-  <summary>Click me to see how it is done</summary>
+Edit file `pyproject.toml` to add the line
 
-  #### Create a virtual environment using python
-
-  ```
-  python -m venv .venv
-  ```
-
-  Or
-
-  ```
-  python -m venv /path/to/new/virtual/environment
-  ```
-
-  Replace `/path/to/new/virtual/environment` with your desired directory path.
-
-  Then activate the environment On linux or MacOS
-
-  ```
-  source /path/to/new/virtual/environment/bin/activate
-  ```
-
-  While on Windows it's
-
-  ```
-  C:\> /path/to/new/virtual/environment/Scripts/activate
-  ```
-
-  Remember to replace `/path/to/new/virtual/environment` with the actual path to your virtual environment. Once the environment is activated, you can clone emu_mps and install it using
-
-</details>
-
-### installing from the registry
-
-When pip is configured to know about the pasqal registry, Emu-MPS installs as
-
-```bash
-pip install emu-mps
-```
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -13,9 +13,9 @@ authors = [
     { name = "Pablo Le Henaff", email="pablo.le-henaff@pasqal.com" }
 ]
 requires-python = ">=3.10"
-license = {text = "Proprietary"}
+license = {text = "MIT-derived"}
```
a-quelle, line 16: this will also need to be done in the `pyproject.toml` files buried in the `ci` directory. There's two of them.

README.md
```diff
@@ -4,110 +4,64 @@
   # Emu-MPS
 </div>
 
-**EMU-MPS** is a Pulser backend, designed to **EMU**late the dynamics of programmable arrays of neutral atoms, with matrix product states (**MPS**). It allows users to increase the number of qubits and reduce computation time.
+**EMU-MPS** is a backend for the [Pulser low-level Quantum Programming toolkit](https://pulser.readthedocs.io). EMU-MPS lets you transparently run Quantum Algorithms on a simulated device, using GPU acceleration if available. More in depth, EMU-MPS is designed to **EMU**late the dynamics of programmable arrays of neutral atoms, with matrix product states (**MPS**). This design makes it faster and more memory-efficient than previous generations of quantum emulators, which means that you can emulate systems with larger number of qubits, faster.
```
a-quelle, line 7: It will be faster above a certain threshold due to exponential scaling of state vector emulators. The above implies that it's faster always, which is definitely not true. Perhaps we should also hedge a bit and compare with state vector emulators specifically.

README.md
```diff
@@ -4,110 +4,64 @@
...
-  pre-commit install
-  ```
-</details>
+For the time being, the easiest way to learn how to use this package is to look
+at the [examples](examples/emu_mps_examples).
 
-## Check the tutorial notebooks and example scripts
+See also the [full API documentation](https://pqs.pages.pasqal.com/emu_mps/).
 
 For more information, you can check the tutorials and examples located in the [examples folder](https://gitlab.pasqal.com/emulation/rydberg-atoms/emu-ct/-/tree/main/examples?ref_type=heads)
```
a-quelle, line 66: this is essentially the same folder as the one you refer to in line 62. This entire sentence can be removed, I think.

README.md
```diff
@@ -4,110 +4,64 @@
...
-  ```
+Edit file `pyproject.toml` to add the line
 
-  While on Windows it's
-
-  ```
-  C:\> /path/to/new/virtual/environment/Scripts/activate
-  ```
+```toml
+  "emu-mps"
```
a-quelle, line 18: I think we should add a warning that this will actually download the package from pypi, not install it editable from source. I don't actually see the point of having this paragraph, since the entire point of hatch is that it installs things in editable mode and manages the environments for you. Why install this package in a hatch env if all it does is give you the pypi version in an env with nothing else?

README.md
```diff
@@ -4,110 +4,64 @@
...
-  ```
+## Usage
 
-  When inside the shell with development dependencies, install first the pre-commit hook:
-  ```
-  pre-commit install
-  ```
-</details>
+For the time being, the easiest way to learn how to use this package is to look
+at the [examples](examples/emu_mps_examples).
```
a-quelle, line 62: also link to the notebooks page in the github pages. This contains rendered notebooks which are not present in the examples.

README.md
```diff
@@ -4,110 +4,64 @@
...
 
-## Check the tutorial notebooks and example scripts
+See also the [full API documentation](https://pqs.pages.pasqal.com/emu_mps/).
```
a-quelle, line 64: this link is broken. Should it link to the github pages page attached to this repo?

README.md
a-quelle: https://pasqal-io.github.io/emulators/latest/

README.md
Yoric: We're trying to standardize on hatch as a package manager, as it saves you from having to manage the virtual env yourself, which is already one less cause of headache.

I don't understand why users would expect it to be editable from source?

README.md
Yoric: Reformulated, what do you think?

README.md
Yoric: Note to self: I'll need to take a look at the contents of `doc/`, too.

README.md
a-quelle: I think I'm a bit unclear what specific workflow you're trying the explain for the user in the case of hatch. 

They can use hatch to create an environment, and then use the `pip install emu-mps` that you describe elsewhere in the README, for which you don't need to checkout the code and change the `pyproject.toml`.

When the user does check out the code, I would expect that they'd want to install that specific version of the emulator, and potentially have it be editable, both of which will not happen if you modify the `pyproject.toml` as you describe.

Maybe there is some alternative use-case that I'm missing somehow, but in the first case, the entire statement about modifying the `pyproject.toml` can be removed, and in the second case, I'd expect some mention of the potentially undesirable behaviour.

Yoric: My scenario is, for instance, [quantum-evolution-kernel](https://github.com/pasqal-io/quantum-evolution-kernel), which depends on emu-mps.

We specify the dependency in `pyproject.toml`. No `pip` involved. No `venv` involved. `hatch run` or `hatch shell` will install the dependencies as needed. Or, for users who prefer `uv` or `poetry`, the local equivalents.

I believe that depending on `emu-mps` will, by far, be the most common use case for `emu-mps`. And I believe that users expect a package they depend on to be downloaded from pypi unless they take the necessary measures to ensure otherwise.

a-quelle (approved):
lgtm

PR #5, opened 2025-01-13
Allow resuming simulation by regularly saving to file

pablolh:
- Adds autosave_prefix and autosave_dt config fields
	By default:
               `emu_mps_save_<uuid>.dat`
		every 10 minute
- Accidentally fix the various config entries and
carry the whole MPSConfig along. It was part of my
research for optimal implementation and takes too
much time to separate as a independent MR.
- Some fixes to the unpickled impl are necessary,
like autosave_file (to save to same existing file)
and the logger which gets messed up by the pickling.
- Fix make_H modifying its interaction_matrix argument.
That was also part of the process but in fact now
unrelated.

---> How to resume a simulation from a file

```
import emu_mps
results = emu_mps.MPSBackend().resume("emu_mps_save_xxx.dat")
```

### DIFF
--- a/emu_mps/mps_config.py
+++ b/emu_mps/mps_config.py
@@ -1,7 +1,7 @@
 from typing import Any
 
 from emu_base import BackendConfig, State
-from emu_mps.utils import DEVICE_COUNT
+from emu_mps.constants import DEVICE_COUNT
 
 
 class MPSConfig(BackendConfig):
@@ -23,6 +23,10 @@ class MPSConfig(BackendConfig):
         num_gpus_to_use: during the simulation, distribute the state over this many GPUs
             0=all factors to cpu. As shown in the benchmarks, using multiple GPUs might
             alleviate memory pressure per GPU, but the runtime should be similar.
+        autosave_prefix: filename prefix for autosaving simulation state to file
+        autosave_dt: minimum time interval in seconds between two autosaves
+            Saving the simulation state is only possible at specific times,
+            therefore this interval is only a lower bound.
         kwargs: arguments that are passed to the base class
 
     Examples:
@@ -43,6 +47,8 @@ class MPSConfig(BackendConfig):
         max_krylov_dim: int = 100,
         extra_krylov_tolerance: float = 1e-3,
         num_gpus_to_use: int = DEVICE_COUNT,
+        autosave_prefix: str = "emu_mps_save_",
+        autosave_dt: int = 600,  # 10 minutes
         **kwargs: Any,
     ):
         super().__init__(**kwargs)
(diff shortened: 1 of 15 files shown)


murogrande:
- is this autosave going to work also with the noise ?

pablolh:
> is this autosave going to work also with the noise ?

It is designed to work with all sorts of noise. I tried it in a bunch of cases, notably the noisy benchmarks were tested for same plots after the previous MR.
Of course I cannot guarantee it is 100% bug-free but I haven't discovered an issue so far.

emu_mps/mps_config.py
```diff
@@ -43,6 +47,8 @@ def __init__(
         max_krylov_dim: int = 100,
         extra_krylov_tolerance: float = 1e-3,
         num_gpus_to_use: int = DEVICE_COUNT,
+        autosave_prefix: str = "emu_mps_save_",
+        autosave_dt: int = 60,
```
a-quelle, line 51: should we default this to `3600` i.e. an hour?

emu_mps/mps_config.py
```diff
@@ -62,3 +68,11 @@ def __init__(
                 and self.noise_model.amp_sigma != 0.0
             ):
                 raise NotImplementedError("Unsupported noise type: amp_sigma")
+
+        self.autosave_prefix = autosave_prefix
+        self.autosave_dt = autosave_dt
+
+        MIN_AUTOSAVE_DT = 10
```
a-quelle, line 75: why this? To avoid some kind of issues with file operations being non-atomic?

a-quelle:
> > > is this autosave going to work also with the noise ?
> > 
> > 
> > It is designed to work with all sorts of noise. I tried it in a bunch of cases, notably the noisy benchmarks were tested for same plots after the previous MR. Of course I cannot guarantee it is 100% bug-free but I haven't discovered an issue so far.
> 
> In this case, then where is stored the number of runs of the monte carlo and in which run the code is actually running? or how is the code managing this situation?

Having several runs for Monte Carlo in one go was something we supported in emu-tn, but here it assumes a single run. If you want to have several runs, you need to write your own code to do it. This code allows you to checkpoint an individual run. You could build a checkpointing system for many runs out of this.

emu_mps/mps_config.py
pablolh: I think more often (10 minutes) is better but the logs will be annoying for very long simulations
What about writing a single log at the beginning? And/or moving the warning log on autosave to a debug/info log

pablolh: The reason why 10 minutes is that I don't think there's a reason for longer, since 
1. pickle is quick
2. if pickle takes a bit of time, it's because there is a lot of data (many gbs), which happens when bond dim is high and simulation steps take hours, therefore it wouldn't happen every 10 minutes
3. outside of crashes, I personally get sometimes problem with internet connection dropping and cluster ssh disconnecting when running simulations in interactive fashion, and waiting an hour can be already quite long (although in that case it's better to run sbatch)

pablolh:
fixed the logging and default autosave time interval

PR #6, opened 2025-01-14
[Emu-SV] Monkey patch and qubit density sparse apply

murogrande:
Related to issue #7 
Monkey patching the apply method of Qubit Density callback. The custom apply is "sparse" implementation of the following operator: 

$I \otimes I \otimes n_i\otimes ... ,$
where $n_i$ is the number operator

New/updates
- emu_sv folder: `custom_applies.py` script where I will store the custom apply methods 
- test: emu-mps: I added a test related to `qubit_density` callback. The same is emu-sv. Both results should be same upto certain tolerance


emu_sv/sv_config.py
```diff
@@ -4,12 +4,20 @@
...
+
+
 from emu_base import BackendConfig
 from emu_sv import StateVector
 from typing import Any
 
 
+from types import MethodType
+
+from emu_sv.custom_applies import custom_qubit_density
```
pablolh, line 18: I would rename `custom_applies` to `custom_callback_implementations`, that is more explicit

emu_sv/custom_applies.py
```diff
@@ -0,0 +1,22 @@
...
+from emu_base.base_classes.operator import Operator
+
+from emu_sv import StateVector
+
+
+def custom_qubit_density(
+    self: QubitDensity, config: BackendConfig, t: int, state: StateVector, H: Operator
+) -> Any:
+
+    num_qubits = int(math.log2(len(state.vector)))
```
pablolh, line 17: Doesn't `StateVector` already store the number of qubits?
`state.qubit_count` would be better

emu_sv/custom_applies.py
```diff
@@ -0,0 +1,22 @@
...
+    state_tensor = state.vector.reshape((2,) * num_qubits)
```
pablolh, line 18: Another idea (maybe for next change) is to provide methods to access either the raw linear vector or the "shaped" tensor, something along the lines of
```
class StateVector:
  def vector_view(self) -> torch.Tensor:
    return self.vector

  def tensor_view(self) -> torch.Tensor:
    return self.vector.reshape((2,) * self.qubit_count)
```

test/emu_mps/test_endtoend.py
```diff
@@ -228,6 +228,11 @@ def test_end_to_end_afm_ring():
     assert fidelity_state.inner(final_state) == approx(final_fidelity, abs=1e-10)
     assert max_bond_dim == 29
 
+    q_density = result["qubit_density"][final_time]
+
+    # 0.578 is the magnetization in the first atom,
```
pablolh, line 233: magnetization of _all_ atoms?

pablolh: Typically, I'm not sure this comment is worth writing as the code is explicit that this is expected value

emu_sv/custom_applies.py
murogrande: No, it does not have it and I think this is a good idea to have. In the future, it will be good to have this for future implementations. 
Is it a bad idea to implement this now? or In a future PR?

test/emu_mps/test_endtoend.py
murogrande: I will delete the comment.

emu_sv/custom_applies.py
pablolh: I think storing an integer is fine

emu_sv/custom_applies.py
pablolh: As you wish, I get quite triggered by this log2 but if it impacts too much I guess this is fine

emu_sv/sv_config.py
```diff
@@ -4,12 +4,20 @@
...
+
+
 from emu_base import BackendConfig
 from emu_sv import StateVector
 from typing import Any
 
 
+from types import MethodType
+
+from emu_sv.ustom_callback_implementations import custom_qubit_density
```
a-quelle, line 18: typo in the file name

emu_sv/sv_config.py
```diff
@@ -66,3 +75,8 @@ def __init__(
         self.max_krylov_dim = max_krylov_dim
         self.gpu = gpu
         self.krylov_tolerance = krylov_tolerance
+
+        for obs in self.callbacks:
+            if isinstance(obs, QubitDensity):
+                # mypy: ignoring dynamically replacing method
+                obs.apply = MethodType(custom_qubit_density, obs)  # type: ignore
```
a-quelle, line 82: Could you specifically ignore only the error it gives? It will make it more explicit what is actually happening.

emu_sv/ustom_callback_implementations.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def custom_qubit_density(
+    self: QubitDensity, config: BackendConfig, t: int, state: StateVector, H: Operator
+) -> Any:
+
+    num_qubits = int(math.log2(len(state.vector)))
+    state_tensor = state.vector.reshape((2,) * num_qubits)
+    return [
+        torch.sum(torch.abs(state_tensor.select(i, 1)) ** 2).item()
```
a-quelle, line 20: It is shorter code, and I expect also faster to call `norm` rather than first `abs` and then `sum`, which will cause two separate loops over the vector. You will still need the square.

emu_sv/ustom_callback_implementations.py
```diff
@@ -0,0 +1,22 @@
...
+import torch
+
+from emu_base.base_classes.config import BackendConfig
+from emu_base.base_classes.default_callbacks import QubitDensity
+from emu_base.base_classes.operator import Operator
+
+from emu_sv import StateVector
+
+
+def custom_qubit_density(
```
a-quelle, line 13: We should also have a unit test for this function. You could create the Hadamard or bell state or something like that using the string format, for which the qubit density is known, and then run this function and assert that the expected answer comes out.

emu_sv/sv_backend.py
```diff
@@ -65,7 +65,7 @@ def run(self, sequence: Sequence, sv_config: BackendConfig) -> Results:
             # TODO: Rydberg Mamiltonian should be a dense operator
 
             H = DenseOperator  # Energy, SecondMomentun... and Variance are not implemented
-
+            # ham :DenseOperator
```
a-quelle, line 68: this seems to be unused

a-quelle (commented):
Apart from my comments on the code, I have an additional question:
Do you want to do the correlation matrix and energy related observables in this MR also, or in a separate one? The approach you take here is good, so doing the others will be kind of the same.

emu_sv/custom_applies.py
murogrande: let me see, if in future PRs, I  need the number of qubits. If this is so, I will add a method.

murogrande:
> Apart from my comments on the code, I have an additional question: Do you want to do the correlation matrix and energy related observables in this MR also, or in a separate one? The approach you take here is good, so doing the others will be kind of the same.

I will do it in a separate MR. I will small code changes that are easy to review

emu_sv/ustom_callback_implementations.py
murogrande: I did not find a function that computes the norm ^2, let me change the code.

emu_sv/sv_config.py
murogrande: Mypy is complaining that I am changing a method dynamically and it expects an immutable method

murogrande: This is the solution: `# type: ignore [method-assign]`

test/emu_sv/test_custom_callbacks.py
```diff
@@ -0,0 +1,24 @@
...
+def test_custom_qubit_density():
+    basis = ("r", "g")
+    num_qubits = 4
+    strings = {"rrrr": 1.0, "ggggg": 1.0}
+    state = StateVector.from_state_string(
+        basis=basis, nqubits=num_qubits, strings=strings
+    )
+
+    config = SVConfig()
+    H = DenseOperator
```
pablolh, line 19: Weird, this H is supposed to be a hamiltonian, not the type of the hamiltonian
You can make this a MagicMock()

murogrande: in principle, H has to be an Operator. Let me mock this

emu_sv/custom_applies.py
pablolh: to do later

test/emu_sv/test_custom_callbacks.py
```diff
@@ -20,16 +20,16 @@ def test_custom_qubit_density():
...
-    MockOperator = MagicMock(spec=DenseOperator)
+    operator_mock = MagicMock(spec=DenseOperator)
 
-    H_mock = MockOperator.return_value
+    H_mock = operator_mock.return_value
 
     MockQubitDensity = MagicMock(spec=QubitDensity)
 
-    qubitDenistyMock = MockQubitDensity.return_value
+    qubit_denisty_mock = MockQubitDensity.return_value
```
pablolh, line 29: qubit_density_mock

pablolh (approved):

PR #11, opened 2025-01-16
Fix test_evolve_pair and test_evolve_single

pablolh:
Thanks to Anton
The operators were transposed

### DIFF
--- a/test/emu_mps/test_tdvp.py
+++ b/test/emu_mps/test_tdvp.py
@@ -196,11 +196,11 @@ def test_evolve_single():
     state_factor = torch.rand(3, 2, 4, dtype=torch.complex128)
     ham_factor = torch.rand(4, 2, 2, 5, dtype=torch.complex128)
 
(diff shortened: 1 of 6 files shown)


test/emu_mps/test_tdvp.py
```diff
@@ -212,16 +212,16 @@ def test_evolve_single():
...
             is_hermitian=False,
-            max_error=1e-10,
-            max_rank=100,  # FIXME: max_error and max_rank are irrelevant for evolve_single
+            max_error=1e-5,
+            max_rank=10,  # FIXME: max_error and max_rank are irrelevant for evolve_single
         ),
     )
 
-    assert torch.allclose(expected, actual, atol=1e-4)
+    assert torch.allclose(expected, actual, atol=1e-8)
```
a-quelle, line 224: an issue I ran into, is that with the default value of `rtol=1e-5` this would pass even for `atol=1e-20` since the relative error is small enough. I expect this problem to exist in other places in the code too, actually. Perhaps you could do a quick search on all `torch.allclose` instances in the tests?

a-quelle (approved):
lgtm

Issue #8, opened 2025-01-15
[Emu-SV] Update: EnergyVariance Callback


murogrande:
**Update the apply method of the EnergyVariance callback**

After issue #6 is merged, we know how to monkey patch the apply method of any callback. 
Now, update the EnergyVariance callback (and others) in order to get the energy variance observable. 

- **The issue**: RydbergHamiltonian class is not an operator thus it does not support operations like `H@H`. This `H@H` is implemented in the apply method of `EnergyVariance`. 
-  Also, the correlation matrix $n_i n_i$ is implemented as a dense operator. 

**The aim of this issue:**  change this behavior for a "sparse" implementation in the case of `EnergyVariance`, `CorrelationMatrix` and `SecondMomentOfEnergy`. In order to accomplish this use the monkey patch.

**Acceptance criteria**
- Finish the issue by updating the `end_to_end` test. Add the energy observable at each time step or at the end and assert the restul - should be the same as in emu-mps. 
-  The RydbergHamiltonian class should be a `DenseOperator`? if it is add the method that it needs in order to be a `DenseOperator`

*Energy*: since `RybergerHamiltonian` does not have an `expect`, we also can not use the energy callback, but if we add an `expect` method the default callback should work.

PR #17, opened 2025-01-21
[Emu-SV]: observables - improve performance

murogrande:
Related to #8 
Performance Improving and bug fixing 
Updating the apply methods of EnergyVariance, SecondMomentOfEnergy and CorrelationMatrix. 
In the case of EnergyVariance, `H@H` is not possible because `RydbergHamiltonian` class is not an operator. The same happens with `SecondMomentOfEnergy`.
In the case of `CorrelationMatrix`, the $n_i n_j$ is implemented as a dense operator. We should make it, sparse.

- Using monkey patch: we have created 'sparse' apply methods for the above observables
- we have added tests for each observable 

Fixing bugs:  

1.  monkey patch makes that emu-mps and emu-sv can not work together. When the apply method is changed, it is changed in emu-base class. 
- Solution: deep copy of the observable 
2. Emu-sv: when using gpu: device='gpu', this does not recognize torch. 
- Solution: instead of 'gpu', use 'cuda'


emu_sv/time_evolution.py
```diff
@@ -18,8 +18,13 @@ def do_time_step(
...
-        op, state_vector, norm_tolerance=krylov_tolerance, exp_tolerance=krylov_tolerance
+    op = lambda x: -1j * dt * (ham * x)
+    return (
+        krylov_exp(
+            op,
+            state_vector,
+            norm_tolerance=krylov_tolerance,
+            exp_tolerance=krylov_tolerance,
+        ),
+        ham,
```
pablolh, line 29: it's weird that you didn't have to change the function signature to be able to do this, is mypy working properly?

test/emu_sv/test_custom_callbacks.py
```diff
@@ -33,3 +51,82 @@ def test_custom_qubit_density():
...
+    config = SVConfig()
+
+    omega = torch.randn(num_qubits, dtype=dtype, device=device)
+    delta = torch.randn(num_qubits, dtype=dtype, device=device)
+    interaction_matrix = torch.randn((num_qubits, num_qubits))
+    h_rydberg = RydbergHamiltonian(
+        omegas=omega, deltas=delta, interaction_matrix=interaction_matrix, device=device
+    )
+
+    MockEnergy = MagicMock(spec=Energy)
```
pablolh, line 107: `energy_mock`

emu_sv/custom_callback_implementations.py
```diff
@@ -15,3 +24,63 @@ def custom_qubit_density(
...
+
+        correlation_matrix.append(one_correlation)
+    return correlation_matrix
+
+
+# feeding RydbergHamiltonian class as an Operator for performance reasons
+def custom_energy(
+    self: Energy, config: BackendConfig, t: int, state: StateVector, H: RydbergHamiltonian
+) -> Any:
+    return torch.vdot(state.vector, H * state.vector).item()
```
pablolh, line 60: I don't fully understand why we need to monkey patch this one?
It's the typing constraints, or?
Same below, it seems you re-wrote the existing default callbacks
The fact that RydbergHamiltonian is not an Operator doesn't mean that it cannot implement an `expect` function for instance, and then you can use the default callbacks no?

emu_sv/time_evolution.py
murogrande: In this case mypy is not complaining. Mypy complains with the monkey Patch, we are changing a method in a class that should be inmutable. I avoid mypy using the `# type: ignore[method-assign]`

emu_sv/custom_callback_implementations.py
murogrande: You are right, let me change it

emu_sv/time_evolution.py
pablolh: yes, exactly, but you didn't have to change the return type, why is that?

pablolh: It's a torch problem, I think
they don't provide any sort of type annotations/hints

emu_sv/custom_callback_implementations.py
```diff
@@ -15,3 +23,56 @@ def custom_qubit_density(
     num_qubits = int(math.log2(len(state.vector)))
     state_tensor = state.vector.reshape((2,) * num_qubits)
     return [(state_tensor.select(i, 1).norm() ** 2).item() for i in range(num_qubits)]
+
+
+def custom_correlation_matrix(
```
sgrava, line 28: "custom" does not add any valuable info, maybe `qubit_density_sv_impl` or something along this line is better

test/emu_sv/test_custom_callbacks.py
```diff
@@ -30,6 +46,80 @@ def test_custom_qubit_density():
...
+    basis = ("r", "g")
+    num_qubits = 4
+    strings = {"rgrg": 1.0, "grgr": 1.0}
+    state = StateVector.from_state_string(
+        basis=basis, nqubits=num_qubits, strings=strings
+    )
+    config = SVConfig()
+    operator_mock = MagicMock(spec=DenseOperator)
+    H_mock = operator_mock.return_value
+    MockCorrelation = MagicMock(spec=CorrelationMatrix)
```
pablolh, line 65: `correlation_matrix_mock`

test/emu_sv/test_hamiltonian.py
```diff
@@ -66,7 +66,7 @@ def test_dense_vs_sparse():
         omegas=omega, deltas=delta, interaction_matrix=interaction_matrix, device=device
     )
 
-    res_sparse = (h_custom @ v).reshape(-1)
+    res_sparse = (h_custom * v).reshape(-1)
```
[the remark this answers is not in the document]
murogrande: This is based on the operator class.  
* : is for operator multiplication a vector
@ : for operator - operator multiplication

pablolh (approved):

PR #21, opened 2025-01-22
Fix mypy with Torch

pablolh:
Add torch as mypy dependency, which revealed lot of new typing errors to be fixed.


emu_base/utils.py
```diff
@@ -1,9 +1,9 @@
 import torch
 
 
-def dist2(left: torch.tensor, right: torch.tensor) -> torch.Tensor:
-    return torch.norm(left - right) ** 2
```
sgrava, line 5: The problem here was caused by `torch.norm`, which is not only deprecated but has this weird `Any | Tensor` signature, right?
`torch.linalg.norm` or `torch.linalg.vector_norm` have better signature, see [here](https://pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm).
With them, you can keep the `**2` and `**3` which are better than `pow`, which is probably intended to do powers with float exponents.

emu_sv/state_vector.py
```diff
@@ -75,23 +78,45 @@ class StateVector(State):
...
+
+        Example:
+        -------
+        >>> StateVector.zero(2)
+        tensor([0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j], dtype=torch.complex128)
+        """
+
+        device = "cuda" if gpu else "cpu"
+        vector = torch.zeros(2**num_sites, dtype=dtype, device=device)
+        return cls(vector, gpu=gpu)
```
sgrava, line 119: While cool, it is dangerous because zero is not a valid state (it is not normalized). Given that making a zero torch tensor is reasonably easy to do, I would avoid having this.

emu_sv/dense_operator.py
```diff
@@ -114,7 +114,7 @@ def expect(self, state: State) -> float | complex:
         ), "currently, only expectation values of StateVectors are \
         supported"
 
-        return torch.vdot(state.vector, self.matrix @ state.vector)  # type: ignore [no-any-return]
+        return torch.vdot(state.vector, self.matrix @ state.vector).item()
```
murogrande, line 117: I remember that @sgrava was against of `.item()` but I do not remember why. Probably, it was for differentiation

emu_sv/state_vector.py
murogrande: I think this `zero` state is important to have. It is useful when you want to use a "zero state" as an accumulator. 
Also, in emu-mps you can have a MPS (which is a state) with all zero tensors. 
It does not make physical sense but from the code point of view is useful

emu_sv/dense_operator.py
sgrava: yeah, everything will return torch stuff in the future, but that doesn't need to be solved here. 
It is already a good amount of work to have solved the mypy complaints :+1:

emu_sv/state_vector.py
sgrava: I see it being useful, but just be careful about that zero **is not** a state.
Besides that `torch.zeros(2**num_sites, dtype=dtype, device=device)` is already convenient, I would say.

emu_sv/dense_operator.py
pablolh: is it okay for differntiability to turn a float back into a tensor?
I don't mind removing the `.item()` but it's less precise in terms of typing - returning a `float | complex` is more precise than a `torch.Tensor`

emu_base/utils.py
pablolh: solved, I also changed other instances of `torch.norm`

emu_sv/state_vector.py
pablolh: I kept the zero factory method but added comments regarding the "physical meaning" if that makes sense

emu_sv/dense_operator.py
pablolh: let's address this later, but good to keep in mind when implementing the new API for emulators

sgrava (approved):

PR #24, opened 2025-01-23
[Emu-SV] tutorial emu-sv and correlation mat performance

murogrande:
New changes: 
- new tutorial of emu-sv. This tutorial is based on Joseph's notebook. What I added are some comments regarding the physics. 
- In my last PR, I changed the `max-line-length` variable from 100 to 110 in the pyproject.toml. I did it because I had to silence mypy using `#type: ignore [method-assing]`, this leads to have line lengths bigger than 100
- There is a way to improve further the correlation matrix by using its symmetry.


a-quelle (approved):
lgtm

PR #25, opened 2025-01-27
Store result times as integers instead of floats

pablolh:


emu_mps/mps_backend_impl.py
```diff
@@ -375,11 +375,14 @@ def save_simulation(self) -> None:
     def fill_results(self) -> None:
         normalized_state = 1 / self.state.norm() * self.state
 
+        current_time_int: int = round(self.current_time)
```
Yoric, line 378: Nit: Perhaps an explanation that you're doing this to ensure that the keys are integers?

Yoric (approved):
Looks good to me, thanks!

emu_mps/mps_backend_impl.py
pablolh: I would like to avoid having this comment since it "should be obvious from the types" but that also makes me think that mypy should have warned us already of this problem, and it didn't, the reason is that there is a "forward reference" in terms of types because otherwise we have a circular import (callback uses config and config uses callback) and mypy doesn't deal with that properly...
After changing that mypy complains about the type being float, and other things which I will fix here as well.
I've already had problems with mypy silently not checking types correctly [here](https://github.com/pasqal-io/emulators/pull/21) with torch, so we should be more careful about this in the future...

Yoric: Ah, if only we knew about a language with higher performance than Python, a stronger type system than Python, and better dependency management :angel: 

More seriously, that was just a minor suggestion, feel free to close this conversation whenever you want!

kbidzhiev (approved):

PR #41, opened 2025-02-14
[emu-sv] torchify custom callbacks

### LINKED ISSUE #40: [emu-sv] return torch from patched callbacks
To allow differentiable workflows, expectation value of observables should be returned as `torch.Tensor` objects.

- change the returned type
- update tests

sgrava:
Resolve #40 

To allow differentiable workflows, expectation value of observables should be returned as `torch.Tensor` objects.
- change the returned type
- update tests (no need `pytest` is smart enough)


a-quelle (approved):
lgtm

PR #42, opened 2025-02-17
improve discretization

a-quelle:
While finishing the emu-sv benchmarks, I was disturbed by the following graph:
emu_sv_afm_state_fidelity_odd
I would expect qubit densities to become more similar to pulser with lower dt, because the assumption of piecewise constant Hamiltonian becomes less discontinuous. Some investigation shows that the problem only occurs for odd dt. I made a change in the discretization logic for odd dt, and now the graph is as follows:
emu_sv_afm_state_fidelity


pablolh (approved):
subtle bug, good fix

PR #45, opened 2025-02-18
pulser core 1.3 + validation

a-quelle:
Make everything depend on pulser-core 1.3.*
Extend the shell script to validate that the dev and published packages depend on the same version.


sgrava (approved):
Nice!

PR #47, opened 2025-02-18
Sg/35 emu mps assumes pulser qubit ids to be integers while they should actually be strings

### LINKED ISSUE #35: emu-mps assumes qubit ids to be integers
Pulser QubitIds are strings (from type annotations) but emu-mps fails working with strings in some places.
Tests are using integers.

Let's fix emu-mps to align on Pulser and use strings, that means fixing e.g. the slm mask feature (qubits in the interaction matrix are currently indexed with their integer id) and fixing the tests to use string qubit ids.

sgrava:
Resolve #35 

Fix warnings in tests

### Changes
- Pulser want to create registers with `prefix="q"`, as it will become the default in the future
- A Register has an attribute `reg.qubit_ids` which is a collection of `QubitID` objects. To map a list of indices and this list of ids
      1. slice `reg.qubit_ids[index]`
      2. get the indices of a Set[QubitID] of an index from `reg.find_indices(set_of_qubits_id)`
- `seq.config_slm_mask()` wants a Sequence[QubitID] (Sequence the abstract class of python not a Pulser.Sequence)

### DIFF
--- a/ci/emu_base/pyproject.toml
+++ b/ci/emu_base/pyproject.toml
@@ -21,7 +21,7 @@ classifiers=[
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
 dependencies = [
-  "pulser-core==1.1.*",
+  "pulser-core==1.3.*",
   "torch==2.5.0"]
 dynamic = ["version"]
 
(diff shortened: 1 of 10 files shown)


a-quelle (commented):
LGTM.

I will leave it up to @kbidzhiev to approve this, since you asked that he also look at it.

kbidzhiev:
clear !

PR #50, opened 2025-02-20
fix indexing error in pulser adapter

a-quelle:
The change in #42 introduced two bugs:
- when dt=1, it attemped to interpolate between the final data point  and the nothingness beyond
- it broke application of the laser waist for odd dt

I fixed the logic, and added a test for `extract_omega_delta_phi` with `dt=1`

### DIFF
--- a/emu_base/pulser_adapter.py
+++ b/emu_base/pulser_adapter.py
@@ -159,23 +159,44 @@ def _extract_omega_delta_phi(
 
     step = 0
     t = (step + 1 / 2) * dt
+    omega_1 = torch.zeros_like(omega[0])
+    omega_2 = torch.zeros_like(omega[0])
 
     while t < max_duration:
-        for q_pos, q_id in enumerate(sequence.register.qubit_ids):
-            omega[step, q_pos] = (
-                locals_a_d_p[q_id]["amp"][math.floor(t)]
-                + locals_a_d_p[q_id]["amp"][math.ceil(t)]
-            ) / 2.0
-            delta[step, q_pos] = (
-                locals_a_d_p[q_id]["det"][math.floor(t)]
-                + locals_a_d_p[q_id]["det"][math.ceil(t)]
-            ) / 2.0
-            phi[step, q_pos] = (
-                locals_a_d_p[q_id]["phase"][math.floor(t)]
-                + locals_a_d_p[q_id]["phase"][math.ceil(t)]
-            ) / 2.0
-        if t in global_times:
-            omega[step] *= waist_factors
+        # The sampled values correspond to the start of each interval
+        # To maximize the order of the solver, we need the values in the middle
+        if math.ceil(t) < max_duration:
+            # If we're not the final step, approximate this using linear interpolation
+            # Note that for dt even, t1=t2
+            for q_pos, q_id in enumerate(sequence.register.qubit_ids):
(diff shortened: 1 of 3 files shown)


emu_base/pulser_adapter.py
```diff
@@ -159,23 +159,44 @@ def _extract_omega_delta_phi(
...
+            # we can reuse omega_1 and omega_2 from before
+            for q_pos, q_id in enumerate(sequence.register.qubit_ids):
+                delta[step, q_pos] = (
+                    3.0 * locals_a_d_p[q_id]["det"][t2] - locals_a_d_p[q_id]["det"][t1]
+                ) / 2.0
+                phi[step, q_pos] = (
+                    3.0 * locals_a_d_p[q_id]["phase"][t2]
+                    - locals_a_d_p[q_id]["phase"][t1]
+                ) / 2.0
+            omega[step] = 0.5 * (3 * omega_2 - omega_1)
```
sgrava, line 199: Can this be negative if omega_2 is 0?

emu_base/pulser_adapter.py
```diff
@@ -159,23 +159,44 @@ def _extract_omega_delta_phi(
...
+                ) / 2.0
+                phi[step, q_pos] = (
+                    locals_a_d_p[q_id]["phase"][t1] + locals_a_d_p[q_id]["phase"][t2]
+                ) / 2.0
+            # omegas at different times need to have the laser waist applied independently
+            if t1 in global_times:
+                omega_1 *= waist_factors
+            if t2 in global_times:
+                omega_2 *= waist_factors
+            omega[step] = 0.5 * (omega_1 + omega_2)
```
sgrava, line 187: I don't immediately see why  
> omegas at different times need to have the laser waist applied independently

this laser modulation is just a time independent factor that only depend on the qubit position

emu_base/pulser_adapter.py
a-quelle: `omega_2 - omega_1` is an approximation to the derivative at the end of the sequence. If it is sufficiently negative, then the extrapolated value will be negative. I did not consider this, maybe we should clamp the value to non-negative.

emu_base/pulser_adapter.py
a-quelle: but it does not occur for local channels. So I'm doing a linear interpolation between two of the samples, each of which either might or might have the laser waist applied.

sgrava: Ok, I now see better that the condition is that t1 or t2 are `global_times`, so associated to a global pulse.
So my question is then if it is possible in the first place to have, for example, t1 in `global_times` and t2 not?

emu_base/pulser_adapter.py
sgrava: It is only really a problem if one tries to build a sequence from negative omegas, which will not happen here.

a-quelle: We discussed, and agreed the best solution would be interpolating to 0 in the final step, rather than doing an extrapolation.

emu_base/pulser_adapter.py
sgrava: yes it is possible!

sgrava (approved):
Good

Issue #54, opened 2025-02-26
Bistring sampling performance


kbidzhiev:
I've found that people use a sampler with $N=10^5 $ or even larger, which takes around 20 seconds for 30–50 qubits. When emu-mps is used in a loop, this part becomes very time-consuming.

pablolh:
https://github.com/pasqal-io/emulators/pull/58

PR #58, opened 2025-03-03
Speed up bitstrings sampling for MPS

pablolh:
Use batch tensor computations


pablolh:
Fixes https://github.com/pasqal-io/emulators/issues/54

emu_mps/mps.py
```diff
@@ -189,11 +189,56 @@ def sample(
...
+
+            batch_outcomes = torch.empty(batch_size, self.num_sites, dtype=torch.bool)
+
+            for qubit, factor in enumerate(self.factors):
+                batched_accumulator = torch.tensordot(
+                    batched_accumulator.to(factor.device), factor, dims=1
+                )
+
+                # Probability of measuring qubit == 0 for each shot in the batch
+                probas = batched_accumulator[:, 0, :].abs().square().sum(dim=1)
```
sgrava, line 215: you probably already thought about it but `linalg.vector_norm` should do the same.

pablolh: it was the original version using norm (not exactly vector_norm) and I changed it because I didn't like that it was computing the norm by taking the square root and we would square it back to get the probas...
I didn't find in torch a builtin doing just the sum of squares of abs, but it's probably just a esthetic concern
what do you think?

pablolh: maybe multiplying element-wise by `.conj()` is faster, I haven't checked (yet)
I will do a micro-benchmark like you did for slicing vs index_add

a-quelle (approved):
lgtm

PR #59, opened 2025-03-05
Kb/emu sv api

kbidzhiev:
moving emu-sv and emu-mps to new pulser `BACKEND` API


emu_base/base_classes/default_callbacks.py
```diff
@@ -48,7 +48,12 @@ def apply(self, config: BackendConfig, t: int, state: State, H: Operator) -> Any
             0.0 if config.noise_model is None else config.noise_model.p_false_neg
         )
 
-        return state.sample(self.num_shots, p_false_pos, p_false_neg)
+        return state.sample(
+            num_shots=self.num_shots,
+            # one_state=None,
```
murogrande, line 53: this can be deleted, I think

emu_sv/dense_operator.py
```diff
@@ -169,31 +176,35 @@ def from_operator_string(
...
 
-            dense_op = total_op_per_qubit[0]
-            for single_qubit_operator in total_op_per_qubit[1:]:
-                dense_op = torch.kron(dense_op, single_qubit_operator)
+            dense_op = reduce(torch.kron, single_qubit_gates)
 
             accum_res += coeff * dense_op
-        return DenseOperator(accum_res)
+
+        # return cls(accum_res, eigenstates=eigenstates)
```
murogrande, line 209: this comment can be deleted

emu_sv/state_vector.py
```diff
@@ -40,6 +43,13 @@ def __init__(
 
         device = "cuda" if gpu and DEVICE_COUNT > 0 else "cpu"
         self.vector = vector.to(dtype=dtype, device=device)
+        self._eigenstates = ["0", "1"]
+
+    @property
+    def n_qudits(self) -> int:
```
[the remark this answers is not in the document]
a-quelle: `pulser-simulation` allows for it, and this API is supposed to be general. We might also do it in the future, we'll have to see.

emu_sv/dense_operator.py
kbidzhiev: removed

emu_base/base_classes/default_callbacks.py
kbidzhiev: I'm going to delete the base_class files

emu_sv/dense_operator.py
```diff
@@ -46,154 +56,148 @@ def __repr__(self) -> str:
...
-                "rr": torch.tensor([[0.0, 0.0], [0.0, 1.0]], dtype=torch.complex128),
+                "gg": torch.tensor([[1.0, 0.0], [0.0, 0.0]], dtype=dtype),
+                "gr": torch.tensor([[0.0, 0.0], [1.0, 0.0]], dtype=dtype),
+                "rg": torch.tensor([[0.0, 1.0], [0.0, 0.0]], dtype=dtype),
+                "rr": torch.tensor([[0.0, 0.0], [0.0, 1.0]], dtype=dtype),
             }
-        elif basis == {"0", "1"}:
-            # operators_with_tensors will now contain the basis for single qubit ops,
-            # and potentially user defined strings in terms of these
+        elif set(eigenstates) == {"0", "1"}:
```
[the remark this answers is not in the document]
kbidzhiev: now it raises `NotImplementedError`

emu_sv/state_vector.py
```diff
@@ -167,62 +187,63 @@ def __rmul__(self, scalar: complex) -> StateVector:
...
-    def norm(self) -> float | complex:
+    def norm(self) -> torch.Tensor:
         """Returns the norm of the state
 
         Returns:
             the norm of the state
         """
-        norm: float | complex = torch.linalg.vector_norm(self.vector).item()
-        return norm
+        nrm: torch.Tensor = torch.linalg.vector_norm(self.vector)
```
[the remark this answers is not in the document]
kbidzhiev: ok

emu_sv/state_vector.py
```diff
@@ -41,13 +47,26 @@ def __init__(
...
+    def n_qudits(self) -> int:
+        """The number of qudits in the state."""
+        nqudits = math.log2(self.vector.reshape(-1).shape[0])
+        return int(nqudits)
+
     def _normalize(self) -> None:
-        # NOTE: use this in the callbacks
-        """Checks if the input is normalized or not"""
-        norm_state = torch.linalg.vector_norm(self.vector)
+        """Normalizes the state vector to ensure it has unit norm.
```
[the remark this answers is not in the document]
kbidzhiev: according to convention norm is always on cpu.
I actually need to move to the device norm and only it

emu_sv/sv_backend.py
```diff
@@ -29,72 +32,99 @@ def run(self, sequence: Sequence, sv_config: BackendConfig) -> Results:
...
+        self.statistics = Statistics(
+            evaluation_times=[t / self.target_times[-1] for t in self.target_times],
+            data=[],
+            timestep_count=nsteps,
+        )
+
+        device = "cuda" if self._config.gpu and DEVICE_COUNT > 0 else "cpu"
+
+        if self._config.initial_state is not None:
+            state = self._config.initial_state
```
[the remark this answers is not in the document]
kbidzhiev: I `.clone()` it  so that `_config.initial_state` field is not modified during the evolution. By the end of the evolution initial state will remain |00.0> or other user feed

emu_mps/mps_backend_impl.py
```diff
@@ -33,6 +37,56 @@
 from enum import Enum, auto
 
 
+class Statistics(Observable):
+    def __init__(
+        self,
+        evaluation_times: typing.Sequence[float] | None,
+        data: list[float],
```
pablolh, line 44: I don't like this `data`, it's very hacky
the definition of `apply` abstractmethod specifically has a `kwargs` so that you don't have to do this. Instead: retrieve for instance a `timestep_duration: float` argument from `kwargs` in `apply`

emu_mps/mps.py
```diff
@@ -187,6 +201,7 @@ def sample(
         Returns:
             the measured bitstrings, by count
         """
+        assert one_state is None or one_state == "r" or one_state == "1"
```
pablolh, line 204: `assert one_state in {None, "r", "1"}`

emu_mps/mps_config.py
```diff
@@ -39,39 +59,108 @@ class MPSConfig(BackendConfig):
...
+                "Warning: The runs and samples_per_run values of the NoiseModel are ignored!"
+            )
+
+    def monkeypatch_observables(self) -> None:
+        obs_list = []
+        for num, obs in enumerate(self.observables):  # monkey patch
+            obs_copy = copy.deepcopy(obs)
+            if isinstance(obs, Occupation):
+                obs_copy.apply = MethodType(  # type: ignore[method-assign]
+                    qubit_occupation_mps_impl, obs
```
pablolh, line 133: I am surprised to see `obs` instead of `obs_copy` on that line. This is actually a bug if any of those custom callbacks were to modify internal state of the observable: see [this](https://stackoverflow.com/questions/46525069/how-is-types-methodtype-used)

pablolh: > The second argument what we're binding the function to. When you call the method object (with e.g. method()), the bound object will be passed into the function as the first argument.

pablolh: We want the custom apply to bind to the observable copy, not the observable itself

emu_mps/results.py
```diff
@@ -0,0 +1,71 @@
...
+
+class MPSResults(Results):
+    def _to_abstract_repr(self) -> dict:
+        d = deepcopy(self.__dict__)
+        for key, value in d["_tagmap"].items():
+            d["_tagmap"][key] = str(value)
+        tmp = {}
+        for key, value in d["_results"].items():
+            tmp[str(key)] = value
+        d["_results"] = tmp
```
pablolh, line 45: using dict comprehension is more readable and doesn't require extra `tmp`:
`d["_results"] = {str(key): value for key, value in d["_results"].items()}`

applies to all loops in this function:
`d["_tagmap"] = {k: str(v) for k, v in d["_tagmap"].items()}`

emu_sv/sv_backend.py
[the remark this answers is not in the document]
a-quelle: your solution will not work, because `state` still points to the same object as in `self._config.initial_state` so in line 66, for example, when `state.vector` is changed again, it will still modify the object that was passed into `self._config`. You have to make sure `state` points to a new object, for example by doing
`state = deepcopy(self._config.initial_state)`in line 58.

emu_sv/state_vector.py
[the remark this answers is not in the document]
a-quelle: but this once again moves data back and forth between the cpu and gpu for nothing. Even though the performance impact is unlikely to be large, this is a bad habit to slip into.

emu_mps/results.py
kbidzhiev: done

emu_mps/mps.py
kbidzhiev: done

emu_sv/state_vector.py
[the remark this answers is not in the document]
kbidzhiev: closing

emu_sv/sv_backend.py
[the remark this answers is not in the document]
kbidzhiev: I see your point, ok

emu_sv/custom_callback_implementations.py
```diff
@@ -1,93 +1,117 @@
...
-    h_squared = torch.vdot(hstate, hstate).real
-    energy = torch.vdot(state.vector, hstate).real
-    energy_variance: torch.Tensor = h_squared - energy**2
-    return energy_variance
+    hstate = hamiltonian * state.vector
+    h_squared = torch.vdot(hstate, hstate)
+    energy = torch.vdot(state.vector, hstate)
+    en_var: torch.Tensor = h_squared - energy**2
+    assert torch.allclose(en_var.imag, torch.zeros_like(en_var.imag), atol=1e-8)
+    return en_var.real
```
a-quelle, line 86: it's better to cast to real early, it saves flops in squaring and adding, although given that this is python, maybe it's outweighed by the fact that you reference one fewer attribute. 

Also, why the assert?

emu_sv/custom_callback_implementations.py
```diff
@@ -1,93 +1,117 @@
...
+    nqubits = state.n_qudits
     correlation = torch.zeros(
-        nqubits, nqubits, dtype=torch.float64, device=state_tensor.device
+        nqubits, nqubits, dtype=torch.float64, device=state.vector.device
     )
 
     for i in range(nqubits):
-        select_i = state_tensor.select(i, 1)
+        select_i = state.vector.view(2**i, 2, -1)
+        select_i = select_i[:, 1]
```
a-quelle, line 56: you do know that line just calls `select_i.select(1,1)` under the hood right? And idem in all the other places where you replaced it. The reshape operation can be elided here.

emu_sv/custom_callback_implementations.py
```diff
@@ -1,93 +1,117 @@
...
 ) -> torch.Tensor:
     """
     Custom implementation of the second moment of energy ❬ψ|H²|ψ❭
     for the state vector solver.
     """
-    hstate = H * state.vector
-    return torch.vdot(hstate, hstate).real
+    hstate = hamiltonian * state.vector
+    en_2_mom: torch.Tensor = torch.vdot(hstate, hstate)
+    assert torch.allclose(en_2_mom.imag, torch.zeros_like(en_2_mom.imag), atol=1e-8)
```
a-quelle, line 102: why?

emu_sv/hamiltonian.py
```diff
@@ -133,19 +133,21 @@ def _create_diagonal(self) -> torch.Tensor:
...
+        return diag.view(-1)
 
     def expect(self, state: StateVector) -> torch.Tensor:
         """Return the energy expectation value E=❬ψ|H|ψ❭"""
         assert isinstance(
             state, StateVector
         ), "Currently, only expectation values of StateVectors are supported"
-        return torch.vdot(state.vector, self * state.vector)
+        en = torch.vdot(state.vector, self * state.vector)
+        assert torch.allclose(en.imag, torch.zeros_like(en.imag), atol=1e-8)
```
a-quelle, line 152: why?

emu_sv/state_vector.py
```diff
@@ -140,7 +159,9 @@ def _index_to_bitstring(self, index: int) -> str:
         """
         Convert an integer index into its corresponding bitstring representation.
         """
-        nqubits = int(math.log2(self.vector.reshape(-1).shape[0]))
+        nqubits = self.n_qudits
+        msg = f"index {index} can not exceed Hilbert space size d**{nqubits}"
+        assert index < 2**nqubits, msg
```
a-quelle, line 164: I dislike these asserts in private code. We are the only users, and we have unit tests. The asserts make everything slow, and they are unlikely to cause any errors that the tests will not.

emu_mps/mps_backend_impl.py
pablolh: oh sorry indeed my mistake. I looked at some subclass `apply` which has kwargs but it's just for shortcutting.
yeah reusing is better, and there is no way to achieve what I said with the pulser code

emu_sv/custom_callback_implementations.py
kbidzhiev: I had in mind some error accumulation in the state with time evolution. Such that positive observables acquire imaginary part.

Now I see it's an overkill. I'm going to remove this unnecessary checks

emu_sv/custom_callback_implementations.py
kbidzhiev: assert is removed. see reply above

a-quelle (approved):
It looks good to the extent I can detect now. Let's merge it and see if anything falls over during the next few days.

PR #62, opened 2025-03-11
add results encoder

a-quelle:
2 of the benchmarks are failing, because emu-sv now returns `torch.Tensor` objects for a few of the results.
These do not serialize by default, so a call to `json.dump` fails. I'm adding an encoder to `Results.dump` so that the serialization works again.


sgrava (approved):
Nice!

PR #63, opened 2025-03-11
support irregular target times

a-quelle:
In both emu-sv and emu-mps the time evolution code now takes a list of target times, and evolves from one target time to the other. The pulser adapter now also outputs a list of target times. Target times are now given as
- all multiples of dt
- augmented with the sequence duration
- augmented with the evaluation times of all the observables

This will make the emulation always emulate exactly the sequence duration, and also removes the issue where observables were not printed if they weren't  evaluated at a multiple of dt.

### DIFF
--- a/test/emu_base/test_adapter.py
+++ b/test/emu_base/test_adapter.py
@@ -369,7 +369,6 @@ def test_extract_omega_delta_phi_dt_1(
     TEST_DURATION = 13
     dt = 1
     target_times = torch.arange(0, TEST_DURATION + 1, dt).tolist()
-    print(target_times)
     sequence.get_duration.return_value = TEST_DURATION
 
     if laser_waist is not None:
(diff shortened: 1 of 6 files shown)


test/emu_base/test_adapter.py
```diff
@@ -364,6 +368,8 @@ def test_extract_omega_delta_phi_dt_1(
     Global pulse: Pulse(RampWaveform(8,10.0,0.0),RampWaveform(8,-10,10),0.2)"""
     TEST_DURATION = 13
     dt = 1
+    target_times = torch.arange(0, TEST_DURATION + 1, dt).tolist()
+    print(target_times)
```
murogrande, line 372: I do not think this print should be here

a-quelle: quite so

kbidzhiev (approved):

PR #64, opened 2025-03-12
small memory optimization in emu-sv

a-quelle:
This makes the program store one state-vector fewer when applying the Hamiltonian. I was bored...


kbidzhiev (approved):
very nice

PR #66, opened 2025-03-17
[Emu-sv] Lindblad noise: Density matrix class

murogrande:
adding a file with DensityMatrix(State) class
- basic operations: constructor for a density matrix (as tensor), inner, sample, etc


emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,111 @@
...
+            num_shots: how many bitstrings to sample
+            p_false_pos: the rate at which a 0 is read as a 1
+            p_false_neg: teh rate at which a 1 is read as a 0
+
+        Returns:
+            the measured bitstrings, by count
+        """
+
+        probabilities = torch.abs(self.matrix.diagonal())
+        probabilities /= probabilities.sum()  # multinomial does not normalize the input
```
pablolh, line 101: this is not needed no? the docs say it uses it as "weights" which means it will do its own business properly
I've also checked

emu_sv/utils.py
```diff
@@ -0,0 +1,10 @@
+import math
+import torch
+
+
+def _index_to_bitstring(vector: torch.Tensor, index: int) -> str:
```
pablolh, line 5: instead of passing a tensor here, pass `nqubits: int` directly: you are using the tensor just to get the shape, which is weird

pablolh: also, rename without the prefix underscore, since it is not private

test/emu_base/math/test_krylov_exp.py
```diff
@@ -35,6 +35,8 @@ def op(x):
         assert x.norm() == pytest.approx(1)  # Check all Lanczos vectors have norm 1.
         return torch.tensordot(m, x, dims=1)
 
+    # take H = 0.5 pi X
```
pablolh, line 38: remove

test/emu_sv/test_density_mat.py
```diff
@@ -0,0 +1,106 @@
...
+    assert density_a.inner(density_b) == 0.0
+
+    n_atoms = 2
+
+    density_c = DensityMatrix(density_bell_state, gpu=gpu)
+
+    assert density_c.inner(density_c) == 1.0
+
+    dummy_mat = torch.zeros((2**n_atoms, 2**n_atoms), dtype=dtype)
+    dummy_mat[0, 0] = 1.0
```
pablolh, line 42: you could add some non-real numbers in the matrices

I also suggest creating separate tests for 1 and 2 atoms

test/emu_sv/test_density_mat.py
```diff
@@ -0,0 +1,106 @@
...
+    state_vector = StateVector.make(n_atoms, gpu=gpu)
+
+    density = DensityMatrix.from_state_vector(state_vector)
+    expected = torch.zeros((2**n_atoms, 2**n_atoms), dtype=dtype, device=device)
+    expected[0, 0] = 1.0
+
+    assert torch.allclose(density.matrix, expected)
+
+    bell_state_vec = 1 / math.sqrt(2) * torch.tensor([1, 0, 0, 1], dtype=dtype)
+    bell_state = StateVector(bell_state_vec, gpu=gpu)
```
pablolh, line 59: I think make separate tests, I read `n_atoms = 3` and after you are using `bell_state` which has only 2 atoms, a bit confusing

test/utils_testing/utils_testing.py
```diff
@@ -220,3 +220,19 @@ def pulser_blackman(duration: float, area: float, phase: float = 0.0):
     seq.add(pi_2, "ch0")
 
     return seq
+
+
+def random_density_matrix(num_qubits):
```
pablolh, line 225: this is good but you are not using it, are you?

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(
+        self,
+        matrix: torch.Tensor,
+        *,
+        gpu: bool = True,
+    ):
+        # NOTE: this accepts also zero matrices.
+
+        device = "cuda" if gpu and DEVICE_COUNT > 0 else "cpu"
+        self.matrix = matrix.to(dtype=dtype, device=device)
```
pablolh, line 25: I think you should add a factory method for creating a density matrix for the |00..0> state, which would directly create the underlying matrix on the right device, like the `StateVector.make` method

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,111 @@
...
+        # NOTE: this is not implemented
+        raise NotImplementedError("Not implemented")
+
+    def __rmul__(self, scalar: complex) -> DensityMatrix:
+        # NOTE: this is not implemented
+        raise NotImplementedError("Not implemented")
+
+    def _normalize(self) -> None:
+        # NOTE: use this in the callbacks
+        """Checks if the input matrix has trace equals to 1.0 or not"""
```
sgrava, line 37: > """Checks if the input matrix has trace equals to 1.0 or not"""

Normalize the density matrix state

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return torch.trace(self.matrix.conj().T @ other.matrix).item()
+
+    @classmethod
+    def from_state_vector(cls, state: StateVector) -> DensityMatrix:
+        """convert a state vector to a density matrix"""
+
+        return DensityMatrix(
+            torch.outer(state.vector, state.vector.conj()), gpu=state.vector.is_cuda
+        )
```
sgrava, line 68: They are forced to stay on the same device to perform the outer operation so the gpu argument is redundant

murogrande: I do not agree with you. The DensityMatrix class always puts the result of outer in gpu. This class is built in a way that always puts the input in gpu. The same happens with the StateVector class. Thus, I would like to specify where I want the density matrix be. 
Just in case, this issue appeared to me meanwhile I was working in the related test. Despite using cpu during the test, the outer product was on gpu (becuase my laptop has gpu). Using the gpu argument, helped me to create the density martrix in cpu.

sgrava: Ah I see, sorry I forgot we default to gpu now :)
Feel free to resolve

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,150 @@
...
+        ), "Other state also needs to be a StateVector"
+        assert (
+            self.matrix.shape == other.matrix.shape
+        ), "States do not have the same number of sites"
+
+        return torch.trace(self.matrix.conj().T @ other.matrix).item()
+
+    @classmethod
+    def from_state_vector(cls, state: StateVector) -> DensityMatrix:
+        """convert a state vector to a density matrix"""
```
sgrava, line 70: > """convert a state vector to a density matrix"""

Return the density matrix ρ=|ψ❭❬ψ| representing the pure state  |ψ❭.

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,150 @@
...
+        self.matrix = matrix.to(dtype=dtype, device=device)
+
+    @classmethod
+    def make(cls, n_atoms: int, gpu: bool = True) -> DensityMatrix:
+        result = torch.zeros(2**n_atoms, 2**n_atoms, dtype=dtype)
+        result[0, 0] = 1.0
+        return cls(result, gpu=gpu)
+
+    def __add__(self, other: State) -> DensityMatrix:
+        # NOTE: this is not implemented
```
sgrava, line 34: I would remove this notes

emu_sv/state_vector.py
```diff
@@ -62,7 +64,7 @@ def zero(cls, num_sites: int, gpu: bool = True) -> StateVector:
             The zero state
```
sgrava, line 64: zero is not a state

pablolh: hi, we already had that discussion before no?
It doesn't mean anything in physics but since you implemented an algebra you can always end up in zero
e.g. by subtracting a state from itself, ...

sgrava: Yeah, sorry, I got confused. Not related to this MR.
I just mean, do not call it a state in the doc string.

test/utils_testing/utils_testing.py
murogrande: Now, I added some tests with this

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+
+        assert isinstance(
+            other, DensityMatrix
+        ), "Other state also needs to be a StateVector"
+        assert (
+            self.matrix.shape == other.matrix.shape
+        ), "States do not have the same number of sites"
+
+        return torch.trace(self.matrix.conj().T @ other.matrix).item()
```
a-quelle, line 73: could you put the `cont().T` on `other`? It makes the `inner` function linear in `self`, anti-linear in `other` just as `StateVector.inner`?

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,184 @@
...
+        # Convert outcomes to bitstrings and count occurrences
+        counts = Counter(
+            [
+                index_to_bitstring(self.matrix.diagonal().shape[0], outcome)
+                for outcome in outcomes
+            ]
+        )
+
+        # NOTE: false positives and negatives
+        return counts
```
a-quelle, line 178: If I understand your comment and the code, the error rates are ignored. I now see that the same is true for `StateVector.sample` could you add, in both methods, an assert that the error rates are `0`?

test/emu_sv/test_density_mat.py
a-quelle: I agree with Pablo, and in addition, please also have `dummy_mat` contain some off-diagonal element (it doesn't need to be hermitian, since you use `conj().T` in the implementation, which makes it work always.). In combination with the complex part, this will test which argument is being `conj().T`d

test/emu_sv/test_state.py
```diff
@@ -9,6 +9,9 @@
 device = "cpu"
 # device= "cuda"
 
+device = "cpu"  # "cuda" if torch.cuda.is_available() else "cpu"
```
a-quelle, line 12: what is the meaning of this? should the commented code replace "cpu"?

emu_sv/density_matrix_state.py
murogrande: yes, I will do it. 
It is weird to me that those arguments are not doing anyting

test/emu_sv/test_density_mat.py
murogrande: Now, I am using the function `random_density_matrix(num_qubits)`. This will add the complex values.

test/emu_sv/test_density_mat.py
murogrande: I reorganize the tests

emu_sv/density_matrix_state.py
murogrande: Done !

emu_sv/density_matrix_state.py
a-quelle: they're there for when we do implement the noise, which is on our todo list. The general API mandates them.

test/emu_sv/test_state.py
a-quelle: A way to work around that, is to always generate random tensors on cpu, and move them to `device` after. It would make the tests a bit better, since running them locally would test gpu behaviour,  but we can do it later.

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,195 @@
...
+        >>> density_bell_state = (1/2* torch.tensor([[1, 0, 0, 1], [0, 0, 0, 0],
+        ... [0, 0, 0, 0], [1, 0, 0, 1]],dtype=torch.complex128))
+        >>> density_c = DensityMatrix(density_bell_state, gpu=False)
+        >>> density_c.inner(density_c)
+        (1+0j)
+        """
+
+        assert isinstance(
+            other, DensityMatrix
+        ), "Other state also needs to be a StateVector"
```
sgrava, line 78: > "Other state also needs to be a StateVector" 

"`other` also needs to be a `DensityMatrix`"

emu_sv/density_matrix_state.py
```diff
@@ -0,0 +1,195 @@
...
+        assert (
+            self.matrix.shape == other.matrix.shape
+        ), "States do not have the same number of sites"
+
+        return torch.trace(self.matrix @ other.matrix.conj().T)
```
sgrava, line 83: would be nice to understand which is faster:
1. torch.trace(A @ B) 
2. (A*B).sum()
3.  tensordot(A, B)

2 or 3 probably

murogrande: the best is ` torch.vdot(self.matrix.flatten(), other.matrix.flatten())`
Screenshot from 2025-04-01 11-58-28

a-quelle (approved):
LGTM, just get @pablolh to resolve his open thread if he agrees.

pablolh (approved):

PR #68, opened 2025-03-20
Fix baths not being updated after quantum jump

pablolh:
This was detected thanks to an assert regarding baths sizes. Post-jump orthogonalization of the state would apply qr decompositions and change state factor sizes (in the presence of very low precision), without right baths being updated.

Originally an issue posted on Slack:
https://pasqalworkspace.slack.com/archives/C07MUV5K7EU/p1742474264332599

The right baths are not recomputed when root-finding, because the hami ltonian doesn't change between iterations (within one timestep). However, a jump implies applying an operator to the state: the right baths need to be updated.

### DIFF
--- a/test/emu_mps/test_endtoend.py
+++ b/test/emu_mps/test_endtoend.py
@@ -27,6 +27,7 @@ from emu_mps import (
 import pulser.noise_model
 
 from emu_mps.mps_backend_impl import MPSBackendImpl
+from emu_mps.tdvp import right_baths
 from test.utils_testing import (
     pulser_afm_sequence_grid,
     pulser_afm_sequence_ring,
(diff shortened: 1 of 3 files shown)


murogrande:
is it possible to add a test?

kbidzhiev (approved):

a-quelle:
I agree with Mauro, any bug that we fix, there should be a unittest that fails before the bug fix, and passes after

a-quelle (changes requested):
unit test

a-quelle (dismissed):
I approve, but @pablolh please review the changes I made. If you agree with them, just merge, otherwise, make a comment and I will fix it

test/emu_mps/test_endtoend.py
```diff
@@ -435,17 +437,97 @@ def test_end_to_end_spontaneous_emission():
...
+        "emu_mps.mps_backend_impl.MPSBackendImpl.get_current_right_bath", autospec=True
+    ) as get_current_right_bath_mock:
+        get_current_right_bath_mock.side_effect = check_baths
+
+        result = simulate(seq, noise_model=noise_model, initial_state=initial_state)
 
     final_time = seq.get_duration()
     final_state = result["state"][final_time]
 
+    print(final_state)
```
murogrande, line 472: delete

pablolh:
lgmt but I don't have right to approve cuz I'm author :(

kbidzhiev (approved):

PR #70, opened 2025-03-24
making inner and norm coherent

kbidzhiev:
`inner`, `norm`, `_normalize` are connected now. 
norm and inner return `float | complex` and it's required by parent class

### DIFF
--- a/emu_sv/state_vector.py
+++ b/emu_sv/state_vector.py
@@ -41,13 +41,35 @@ class StateVector(State):
         device = "cuda" if gpu and DEVICE_COUNT > 0 else "cpu"
         self.vector = vector.to(dtype=dtype, device=device)
 
+    def _inner(self, other: State) -> torch.Tensor:
+        """
+        helper function. Computes <self, other>. The type of other must be StateVector.
+
+        Args:
+            other: the other state
+
+        Returns:
+            the inner product
+        """
+        assert isinstance(
+            other, StateVector
+        ), "Other state also needs to be a StateVector"
+        assert (
+            self.vector.device == other.vector.device
+        ), "States are not on the same device"
+        assert (
+            self.vector.shape == other.vector.shape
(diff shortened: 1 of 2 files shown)


emu_sv/state_vector.py
```diff
@@ -102,14 +124,7 @@ def inner(self, other: State) -> float | complex:
...
         """
-        assert isinstance(
-            other, StateVector
-        ), "Other state also needs to be a StateVector"
-        assert (
-            self.vector.shape == other.vector.shape
-        ), "States do not have the same number of sites"
-
-        return torch.vdot(self.vector, other.vector).item()
+        return self._inner(other).cpu().item()
```
a-quelle, line 127: this should return a `torch.Tensor` too, so it's essentially what you have in `_inner` now.

emu_sv/state_vector.py
```diff
@@ -41,13 +41,35 @@ def __init__(
...
+        assert (
+            self.vector.shape == other.vector.shape
+        ), "States do not have the same number of sites"
+        return torch.vdot(self.vector, other.vector)
+
     def _normalize(self) -> None:
         # NOTE: use this in the callbacks
         """Checks if the input is normalized or not"""
-        norm_state = torch.linalg.vector_norm(self.vector)
+        norm_state = self._inner(self)
```
a-quelle, line 68: I'd expect using the inbuilt norm function to be faster here.

emu_sv/state_vector.py
```diff
@@ -41,13 +41,35 @@ def __init__(
         device = "cuda" if gpu and DEVICE_COUNT > 0 else "cpu"
         self.vector = vector.to(dtype=dtype, device=device)
 
+    def _inner(self, other: State) -> torch.Tensor:
```
a-quelle, line 44: who not move this to `inner` and have a dedicated `_norm` for where the norm is needed?

PR #76, opened 2025-03-28
fix warning logic

a-quelle:
The logic for throwing the warning about `noise_model` parameters in the config was broken. Fixed it.


sgrava (approved):
noice!

PR #79, opened 2025-04-01
Kb/77 serialisation tests

kbidzhiev:
- [x] verifies `pulser-core` version, temporarily it was suppressed
- [x] improved readability for script `./test_dependency_versions.py`
- [x] complementary tests for serialisation
- [x] Not allowed to add unsupported keyword for `SV/MPSconfig` and Allowed for `EmulationConfig`
- [x] Default observable for `SV/MPSConfig` is `Bitstrings`. `StateResults` are forbidden by json.serialisation


emu_sv/sv_config.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         log_file: pathlib.Path | None = None,
         **kwargs: Any,
     ):
-        kwargs.setdefault("observables", [StateResult(evaluation_times=[1.0])])
+        kwargs.setdefault("observables", [BitStrings(evaluation_times=[1.0])])
         super().__init__(**kwargs)
         self._backend_options["dt"] = dt
         self._backend_options["max_krylov_dim"] = max_krylov_dim
```
sgrava, line 65: you can just pass them to the super init
```
super._init_(dt=dt, max_krylov_dim=max_krylov_dim,...)
```

kbidzhiev: thanks for pointing that out. I've noticed that guard
`_expected_kwargs` doesn't work and didn't spot what happened

emu_mps/mps_config.py
```diff
@@ -122,9 +122,19 @@ def __init__(
                 "Warning: The runs and samples_per_run values of the NoiseModel are ignored!"
             )
 
+    def _expected_kwargs(self) -> set[str]:
```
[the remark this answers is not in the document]
kbidzhiev: it's not called at all. Now I can fully remove this guard `_expected_kwargs ` and still be able to create `SVConfig()`

emu_sv/sv_config.py
kbidzhiev: done

test/emu_mps/test_mps_config.py
```diff
@@ -34,3 +45,54 @@ def test_unsupported_noise():
...
+def test_serialise_config() -> None:
+    # Arguments are arbitrary just to be != default, no deep meaning behind
+    dt = 1
+    precision = 1e-2
+    max_bond_dim = 10
+    max_krylov_dim = 15
+    extra_krylov_tolerance = 1e-1
+    num_gpus_to_use = 0
+
+    default_config = MPSConfig(
```
sgrava, line 73: I suggest adding the same test but also for `default_config=EmulationConfig(`
the idea is to test that you can also define a correct `MPSConfig` from a serialized `EmulationConfig`.
The same double test also for SVconfig

emu_mps/mps_config.py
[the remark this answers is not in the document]
kbidzhiev: hehe, done and very elegantly. I've added 
`_enforce_expected_kwargs: ClassVar[bool] = True` that works for SvConfig and MPSconfig, but doesn't for Emulationconfig. Tests prove the point

test/emu_mps/test_mps_config.py
kbidzhiev: ok, got the point

kbidzhiev: done

test/emu_sv/test_end_to_end.py
```diff
@@ -116,15 +116,6 @@ def simulate(
     return result
 
 
-def test_default_SVConfig_ctr() -> None:
```
kbidzhiev, line 119: Moved to test_svconfig

emu_mps/mps_config.py
```diff
@@ -149,7 +170,7 @@ def monkeypatch_observables(self) -> None:
             obs_list.append(obs_copy)
         self.observables = tuple(obs_list)
 
-    def init_logging(self) -> None:
+    def init_logging(self) -> Any:
```
sgrava, line 173: why Any?

kbidzhiev: because function returns

a-quelle: no, it does not. Otherwise mypy would've complained. The reason it does not do so now, is because `None` is an `Any`, so the type hint is still valid, but much too general.

test/emu_mps/test_mps_config.py
```diff
@@ -34,3 +50,124 @@ def test_unsupported_noise():
             p_false_neg=0.15,
         )
     )
+
+    pass
+
+
+def test_serialise_default_config() -> None:
```
sgrava, line 57: `test_default_config_repr` I think is better because you are testing serialization and deserialization workflow here, so that the representation is correct

test/emu_mps/test_mps_config.py
```diff
@@ -34,3 +50,124 @@ def test_unsupported_noise():
...
+
+    deserialized_config = MPSConfig.from_abstract_repr(config_str)
+
+    for attr in mps_attributes:
+        assert getattr(deserialized_config, attr) == getattr(
+            default_config, attr
+        ), f"{attr} mismatch"
+
+
+def test_serialise_config() -> None:
```
sgrava, line 69: test_config_repr

test/emu_mps/test_mps_config.py
```diff
@@ -34,3 +50,124 @@ def test_unsupported_noise():
...
+
+    # MPSConfig
+    msg = (
+        "'MPSConfig' received unexpected keyword arguments: "
+        "{'blabla'}; only the following keyword arguments "
+        "are expected:"
+    )
+    # re.escape() avoid interpreting message symbols "{}"", "()","."" as regex
+    with pytest.warns(UserWarning, match="^" + re.escape(msg)):
+        MPSConfig(blabla=10)
```
sgrava, line 173: - `BackendConfig` should not be tested here
- `EmulationConfig`: rather than not catching a warning I would just test the expected behavior. This is tested in pulser so I would remove this

The third one is a test on the expected_kwargs patch.
I would call it `test_expected_kwargs`.
Overall I would just replace all with:
```
def test_expected_kwargs() -> None:
    with pytest.warns(UserWarning, match="received unexpected keyword arguments"):
        MPSConfig(blabla=10)
```

sgrava: On top of that, I would prefer an error rather than a warning

test/emu_mps/test_mps_config.py
```diff
@@ -1,12 +1,28 @@
...
+# copypaste mps_config.py specific attributes
+# attributes from MPSConfig._expected_kwargs
+mps_attributes = [
+    "dt",
+    "precision",
+    "max_bond_dim",
+    "max_krylov_dim",
+    "extra_krylov_tolerance",
+    "num_gpus_to_use",
+]
```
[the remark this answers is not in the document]
kbidzhiev: indeed, thanks

test/emu_mps/test_mps_config.py
kbidzhiev: done

test/emu_mps/test_mps_config.py
kbidzhiev: unified 2 cases. 

Emulationconfig and SV/MPSconfig  serialisation

test/emu_mps/test_mps_config.py
kbidzhiev: The test is designed to demonstrate for users/devs that there are 3 different behaviours.

I do it exactly to capture that `MPS/SVConfig` **does** verify kwargs, while `EmulationConfig` does **not**. It is inexplicit knowledge. 
Before PR we didn't notice that behaviour was changed. 

Though I don't strongly insist to have all 3 here, I think they are complementary demonstrative

kbidzhiev: changing it to error requires design work on pulser side. Or some bigger work on our side.

sgrava: one option could be
```
        if invalid_kwargs := set(kwargs) - super()._expected_kwargs():
            raise TypeError(
                f"MPSConfig received unexpected keyword arguments: "
                f"{invalid_kwargs}; only the following keyword "
                f"arguments are expected: {self._expected_kwargs()}. "
            )
```

because kwargs should only be valid keywords for EmulationConfig

a-quelle: Testing `EmulationConfig` causes the tests to no longer be "unit tests" in the sense that they start testing things outside of this repo. I would just test `MPS/SVConfig` in our tests, and leave responsibility for `EmulationConfig` to pulser.

a-quelle: Ok, so I convinced Henrique that we should just raise an error on the pulser side. So let's leave it at a warning for now, and then change when that gets released in pulser. I still think we should not test `BackendConfig` and `EmulationConfig` in this repo though, that is pulser's responsibility.

sgrava (approved):

PR #80, opened 2025-04-01
[Emu-SV] krylov method for evolving a density matrix

murogrande:
Adding a test for `krylov_exp` function:
I would like to have the a test for $e^{op} |\psi><\psi| $, where is given by`op` $= \frac{1}{2} \Omega \sigma_x \otimes I + \frac{1}{4} \Omega I \otimes \sigma_y$ and the state $|\psi><\psi|$ is the Identity operator $I \otimes I$

The take away is that our `krylov_exp` implementation can exponentiate a matrix and multiply it with another matrix

- update: Since, I am testing the `krylov_exp` can evolve matrices, I will replace the `op` for a random operator and the state will be a random matrix.


test/emu_base/math/test_krylov_exp.py
```diff
@@ -185,23 +188,80 @@ def test_converged_non_hermitian_non_normalized():
     )
 
 
+def test_krylov_with_matrix():
+
```
kbidzhiev, line 192: you use matrix sigma_X and others several times.
You can define it one in the beginning only and reuse it

test/emu_base/math/test_krylov_exp.py
```diff
@@ -185,23 +188,51 @@ def test_converged_non_hermitian_non_normalized():
...
+    const = 3.14159 / 2
+    H = const * torch.kron(sigma_x, Ident) + 1.5 * const + torch.kron(Ident, sigma_y)
+
+    def op(x):
+        return H @ x
+
+    M = torch.tensor([[1.0, 0.0], [0.0, 1.0]], dtype=dtype)
+    MxM = torch.kron(M, M)  # matrix to evolve
+
+    result = krylov_exp_to_matrix(
```
a-quelle, line 206: The test is precisely what I had in mind, however it will also pass if you call `krylov_exp` here. The new implementation you wrote is not needed.

murogrande: indeed, let me eliminate the new function.

test/emu_base/math/test_krylov_exp.py
```diff
@@ -185,23 +188,51 @@ def test_converged_non_hermitian_non_normalized():
     )
 
 
+def test_krylov_with_matrix():
+
+    Ident = torch.eye(2, dtype=dtype)
```
pablolh, line 193: variable names should be lowercase

test/emu_base/math/test_krylov_exp.py
```diff
@@ -185,23 +188,51 @@ def test_converged_non_hermitian_non_normalized():
...
+    const = 0.001 * 3.14159 / 2
+    Ht = const * torch.kron(sigma_x, Ident) + 0.5 * const + torch.kron(Ident, sigma_y)
+
+    def op(x):
+        return Ht @ x
+
+    M = torch.tensor([[1.0, 0.0], [0.0, 1.0]], dtype=dtype)
+    MxM = torch.kron(M, M)  # matrix to evolve
+
+    result = krylov_exp(
```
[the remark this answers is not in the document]
murogrande: what do think is the best? 
- leave it or create a random matrix as Hamiltonian

test/emu_base/math/test_krylov_exp.py
```diff
@@ -185,23 +188,51 @@ def test_converged_non_hermitian_non_normalized():
     )
 
 
+def test_krylov_with_matrix():
```
kbidzhiev, line 191: add the None return type to please mypy

test/emu_base/math/test_krylov_exp.py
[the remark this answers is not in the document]
pablolh: you need to be careful that it matches the `is_hermitian` parameter of `krylov_exp`
it needs to be hermitian as linear operator on matrix space, I don't know how that translates (probably trivial?)
Instead of random, you can hardcode a simple one in low dimension...

test/emu_base/math/test_krylov_exp.py
murogrande: I installed mypy app in vscode and mine is not complaining. Which mypy did you installed?

test/emu_base/math/test_krylov_exp.py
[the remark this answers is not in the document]
kbidzhiev: `M = torch.rand(2**n_atoms, 2**n_atoms, dtype=dtype)`
`MxM = torch.conj(M) @ M`

saves you to have hermitean

test/emu_base/math/test_krylov_exp.py
kbidzhiev: I'm using 1.15. Please consider my comment per your convenience. It's a stylistic thing.

pablolh: ? I thought mypy didn't check that in tests, which makes sense

pablolh (approved):

pablolh:
for me the test is good as proof of concept that krylov_exp works with any op
we can investigate the is_hermitian "issue" separately

PR #81, opened 2025-04-07
Permute observables with optimatrix

kbidzhiev:
done in other PR #94


emu_mps/mps_backend.py
```diff
@@ -54,8 +60,26 @@ def run(self) -> Results:
 
         impl = create_impl(self._sequence, self._config)
         impl.init()  # This is separate from the constructor for testing purposes.
+        results = self._run(impl)
+        if not self._config.optimise_interaction_matrix:
+            return results
+
+
+        inv_perm = opmat.invert_permutation(impl.opt_perm)
```
pablolh, line 68: I would put lines 68 -> 80 into an auxiliary function

pablolh: what if the user didn't ask for storing bitstrings in results?

pablolh: what about
- computing `self.inv_perm` and storing it as class member
- doing this inverse permutation of bitstrings on-the-fly in `fill_results`?

pablolh: shouldn't you ALSO permute the state and hamiltonian that are generically passed to callbacks?
bitstrings are not even the only order-dependent default observable so far: you also have qubit occupation!

emu_mps/mps_backend_impl.py
```diff
@@ -165,6 +170,27 @@ def __setstate__(self, d: dict) -> None:
...
+            m_mat: torch.Tensor,
+            optimise: bool = True
+            ) -> None:
+        if not optimise:
+            self.full_interaction_matrix = f_mat
+            self.masked_interaction_matrix = m_mat
+            return
+
+        f_mat_numpy = f_mat.numpy()
+        m_mat_numpy = m_mat.numpy()
```
pablolh, line 185: An idea is to always store in this class `opt_perm` and rename it to `qubit_permutation` which would default to the identity.
Then, you are only doing
`self.qubit_permutation = optmat.identity(self.num_sites) if not self._config.optimize_qubit_ordering else optimatrix.minimize_bandwidth(self.full_interaction_matrix)`

You would apply the identity permutation to stuff when optmat is disabled, but trust me that performance cost is very small while the code is nicer IMO

pablolh (commented):
my biggest concern is that you are only modifying bitstrings. but for instance **qubit occupation** needs treatment as well. Given that users can define their own callbacks based on state + hamiltonian, I think a better way is to give a permuted view of state and hamiltonian to callbacks. What do you think?

PR #84, opened 2025-04-10
Implementing the Entanglement Entropy in emu-mps

elmerhej:
Implementing an Entanglement entropy observable in emu-mps. To do that I
- Added an entanglement entropy method
- Wrote an EntanglementEntropy subclass that inherets from Observable
- Wrote 3 tests


test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,117 @@
...
+        S_E_res = result.get_result(ent_entropy, time=1.0)
+        entropies.append(S_E_res)
+
+    # Check that central bonds have higher entropy
+    center_bonds = [4, 5]
+    center_entropies = [entropies[b - 1] for b in center_bonds]
+    edge_entropies = [entropies[0], entropies[-1]]
+
+    # Assert that center bonds have entropy > 1e-1
+    assert any(
```
kbidzhiev, line 109: I think you can assert here against analytical results. At least 2 qubit system

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
+from emu_mps.mps import MPS
+import torch
+
+
+def calculate_entanglement_entropy(mps: MPS, b: int) -> float:
+    """
+    Calculate the von Neumann entanglement entropy at the bond between sites b-1 and b
+    S_E = -Tr(s^2 * log(s^2)), where s are the singular values at the chosen bond
+    """
```
kbidzhiev, line 9: why do you say `b-1` ? I see that you do `svd` only on the orthocenter

test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,117 @@
...
+    seq = create_constant_pulse_sequence()
+
+    backend = emu_mps.MPSBackend(sequence=seq, config=config)
+    result = backend.run()
+
+    result_times = result.get_result_times(observable=ent_entropy)
+    assert 0.5 in result_times
+
+    entropy = result.get_result(observable=ent_entropy, time=1)
+    assert isinstance(entropy, float)
```
kbidzhiev, line 52: You test here that 0.5 is in evaluation times and that entanglement entropy is a float? I think it's a bit redundant. 

You can omit the part with `assert 0.5 in result_times`
and put `assert isinstance(entropy, float)` to the other function.

test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,117 @@
...
+    for b in range(1, 9):
+        ent_entropy = EntanglementEntropy(bond_index=b, evaluation_times=[1 / 100])
+        config = emu_mps.MPSConfig(observables=[ent_entropy], initial_state=initial_state)
+        seq = create_constant_pulse_sequence()
+        backend = emu_mps.MPSBackend(sequence=seq, config=config)
+        result = backend.run()
+        S_E_res = result.get_result(ent_entropy, time=1 / 100)
+        entropies.append(S_E_res)
+
+    assert all(s < 1e-7 for s in entropies)
```
kbidzhiev, line 72: this somehow implies that entropy is a float, because you comparing it with floats. I would just remove the test above

test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def create_simple_superposition_state():
+    """
+    Create the max entangled state (|rrrrrrrrr> + |ggggggggg>)/sqrt(2)
+    """
+    return {"rrrrrrrrr": 1 / 2**0.5, "ggggggggg": 1 / 2**0.5}
```
kbidzhiev, line 79: maybe smth like 

`"r" * 9` instead of `"rr..r"` will be easier to read

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+        tensor_at_site_b.shape[0] * tensor_at_site_b.shape[1], tensor_at_site_b.shape[2]
+    )
+
+    _, s, _ = torch.linalg.svd(matrix, full_matrices=False)
+
+    # Calculate entropy from singular values 's'
+    s_sqrd = s**2
+    s_sqrd = s_sqrd[s_sqrd > 1e-12]
+
+    S_E = -torch.sum(s_sqrd * torch.log(s_sqrd))
```
pablolh, line 25: variable names should be lowercase

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Calculate the von Neumann entanglement entropy at the bond between sites b-1 and b
+    S_E = -Tr(s^2 * log(s^2)), where s are the singular values at the chosen bond
+    """
+    # orthogonalize the MPS at site b
+    mps.orthogonalize(b)
+
+    # perform svd on reshaped tensor at index 'b' of the mps
+    tensor_at_site_b = mps.factors[b]
+    matrix = tensor_at_site_b.reshape(
```
pablolh, line 15: a shorter alternative using parameters to [`flatten`](https://pytorch.org/docs/stable/generated/torch.flatten.html)
`matrix = mps.factors[b].flatten(end_dim=1)`

emu_mps/observables.py
```diff
@@ -0,0 +1,40 @@
...
+        repr = super()._to_abstract_repr()
+        repr["bond_index"] = self.bond_index
+        return repr
+
+    def apply(self, *, state: State, **kwargs: Any) -> float:
+        if not isinstance(state, MPS):
+            raise NotImplementedError(
+                "Entanglement entropy observable is only available for emu_mps emulator."
+            )
+        if not (0 < self.bond_index < len(state.factors)):
```
pablolh, line 35: Can you compute the entropy for bond i from MPS tensors of both sides?
If so what about indexing bonds in [[0, qubit_count - 2]]?
It feels more natural from computer perspective (Python lists indexing is 0-based), but maybe debatable from physics perspective...

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+        tensor_at_site_b.shape[0] * tensor_at_site_b.shape[1], tensor_at_site_b.shape[2]
+    )
+
+    _, s, _ = torch.linalg.svd(matrix, full_matrices=False)
```
sgrava, line 19:
```suggestion
    _, s, _ = torch.linalg.svd(matrix, full_matrices=False)
```
torch.linalg.svdvals

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
+from emu_mps.mps import MPS
+import torch
+
+
+def calculate_entanglement_entropy(mps: MPS, b: int) -> float:
```
sgrava, line 5: I suggest this to be a method of the MPS class, to allow `MPS.entanglement_entropy(site: int)`

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Calculate the von Neumann entanglement entropy at the bond between sites b-1 and b
+    S_E = -Tr(s^2 * log(s^2)), where s are the singular values at the chosen bond
```
sgrava, line 8:
```suggestion
    Returns the Von Neumann entanglement entropy of the state `mps` at the bond between sites b-1 and b
         S = -Σᵢsᵢ² log(sᵢ²)), 
    where sᵢ are the singular values at the chosen bond.
```
or something along this line

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    # orthogonalize the MPS at site b
+    mps.orthogonalize(b)
+
+    # perform svd on reshaped tensor at index 'b' of the mps
+    tensor_at_site_b = mps.factors[b]
+    matrix = tensor_at_site_b.reshape(
+        tensor_at_site_b.shape[0] * tensor_at_site_b.shape[1], tensor_at_site_b.shape[2]
```
kbidzhiev, line 16: Elie, you are reshaping and join indices (0,1). Are you sure that bipartition goes between (b,b-1) or is it (b,b+1) ?

elmerhej: Yes you're correct I should have edited the docstrings. The takeway is when the orthogonality center is at site 'b':

- if tensor (left index, physical index, right index) is reshaped into (left index * physical index, right index), then the partition goes between (b, b+1)
- if tensor (left index, physical index, right index) is reshaped into (left index, physical index* right index), then the partition goes between (b-1, b)

emu_mps/custom_callbacks.py
pablolh: I let you and the other choose whatever you prefer, it was just a suggestion :) .
My opinion is that if there is a specialized torch function for sth, better use it.

emu_mps/custom_callbacks.py
a-quelle: yes, let's delete this file, and a method in `MPS` as @sgrava suggests

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # Calculate entropy from singular values 's'
+    s_sqrd = s**2
+    s_sqrd = s_sqrd[s_sqrd > 1e-12]
+
+    S_E = -torch.sum(s_sqrd * torch.log(s_sqrd))
+
+    # shift the orthogonality center back to 0 before the new tdvp step
+    mps.orthogonalize(0)
+    return S_E.item()
```
a-quelle, line 29: for consistency, maybe we should return a `torch.tensor` just as with the other observables?

a-quelle (changes requested):
Apart from the comments I wrote on the code, please export the new observable in the package `__init__.py` so it can be imported conveniently by the user.

emu_mps/observables.py
a-quelle: see also my comment above about the range of `b` values actually permitted in the code

emu_mps/custom_callbacks.py
```diff
@@ -0,0 +1,29 @@
...
+    tensor_at_site_b = mps.factors[b]
+    matrix = tensor_at_site_b.reshape(
+        tensor_at_site_b.shape[0] * tensor_at_site_b.shape[1], tensor_at_site_b.shape[2]
+    )
+
+    _, s, _ = torch.linalg.svd(matrix, full_matrices=False)
+
+    # Calculate entropy from singular values 's'
+    s_sqrd = s**2
+    s_sqrd = s_sqrd[s_sqrd > 1e-12]
```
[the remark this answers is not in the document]
elmerhej: this is just a numerical cutoff to exclude values of the square of the entanglement entropy that are effectively zero, since log (0) --> -inf, so gives something like NaN and instability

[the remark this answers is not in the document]
a-quelle: good point, I forgot you were taking the log later.

test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,102 @@
...
+        eigenstates=("r", "g"), amplitudes={"rrrrrrrrr": 1.0}
+    )
+    entropies = []
+    for b in range(8):
+        ent_entropy = EntanglementEntropy(mps_site=b, evaluation_times=[10 / 100])
+        config = emu_mps.MPSConfig(observables=[ent_entropy], initial_state=initial_state)
+        seq = create_constant_pulse_sequence()
+        backend = emu_mps.MPSBackend(sequence=seq, config=config)
+        result = backend.run()
+        S_E_res = result.get_result(
```
kbidzhiev, line 47: variable name should start with lowercase

emu_mps/mps.py
```diff
@@ -307,6 +307,30 @@ def overlap(self, other: State, /) -> torch.Tensor:
...
+
+        # perform svd on reshaped matrix at site b
+        matrix = self.factors[mps_site].flatten(end_dim=1)
+        s = torch.linalg.torch.linalg.svdvals(matrix)
+
+        # Calculate entropy from singular values 's'
+        s_sqrd = s**2
+        s_sqrd = s_sqrd[s_sqrd > 1e-12]
+
+        s_e = -torch.sum(s_sqrd * torch.log(s_sqrd))
```
sgrava, line 328:
```suggestion
        s = torch.linalg.svdvals(matrix)
        s_e = torch.special.entr(s**2)
```
see [torch.special.entr](https://pytorch.org/docs/stable/special.html#torch.special.entr)

kbidzhiev: so nice, I didn't know about the entropy function

test/emu_mps/test_entanglement_entropy.py
```diff
@@ -0,0 +1,102 @@
...
+
+def test_zero_entropy_product_state():
+    """
+    starting from a product state, the S_E should be zero across all bonds
+    """
+    initial_state = emu_mps.MPS.from_state_amplitudes(
+        eigenstates=("r", "g"), amplitudes={"rrrrrrrrr": 1.0}
+    )
+    entropies = []
+    for b in range(8):
```
pablolh, line 41: why simulate the same sequence 8 times?
You should be able to add all 8 observables to a single simulation. I think that is related to Anton's comments this morning at standup

elmerhej: That's actually a great improvement to the test! I implemented it!

pablolh (commented):
lgtm I'll let others approve

emu_mps/mps.py
```diff
@@ -307,6 +307,25 @@ def overlap(self, other: State, /) -> torch.Tensor:
...
+
+        # perform svd on reshaped matrix at site b
+        matrix = self.factors[mps_site].flatten(end_dim=1)
+        s = torch.linalg.svdvals(matrix)
+
+        s_e = torch.Tensor(torch.special.entr(s**2))
+        s_e = torch.sum(s_e)
+
+        self.orthogonalize(0)
+        return s_e
```
a-quelle, line 327: please call `.cpu()` on the return value. we made a conscious decision to have all api objects on cpu to avoid errors for the user where some things might be on gpu, and others not.

a-quelle (approved):
lgtm

PR #86, opened 2025-04-16
Refactor hamiltonian.py

pablolh:
- base class for MPO hamiltonian factors
- factory method make_H
- no more a posteriori factor truncation
- thanks to objects, fewer parameters passing
- group operators into class/namespace

### DIFF
--- a/emu_mps/hamiltonian.py
+++ b/emu_mps/hamiltonian.py
@@ -3,309 +3,345 @@ This file deals with creation of the MPO corresponding
 to the Hamiltonian of a neutral atoms quantum processor.
 """
 
+from abc import abstractmethod, ABC
+from typing import Iterator
+
 from emu_base import HamiltonianType
 import torch
-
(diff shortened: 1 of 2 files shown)


emu_mps/hamiltonian.py
```diff
@@ -3,309 +3,351 @@
...
-    """
-    fac = torch.zeros(1, 2, 2, 3 if interaction else 2, dtype=dtype)
-    fac[0, :, :, 1] = iden_op
-    if interaction:
-        fac[0, :, :, 2] = n_op  # number operator
 
-    return fac
 
+class Operators:
+    id = torch.eye(2, 2, dtype=dtype)
```
kbidzhiev, line 17: `id = torch.eye(2, dtype=dtype)` does the same

pablolh: done

a-quelle (approved):
lgtm

PR #87, opened 2025-04-16
fix for failing observables after autosave

a-quelle:
The autosave function modifies the results object in place to a `dict`, which causes observables called after the first autosave to fail since the results are no longer of type `Results`.

The fix is to make sure the `__getstate__` function does not modify the object in place.

I added a test `test_obs_after_autosave` that failed before the test was implemented, and now passes.

### DIFF
--- a/emu_mps/mps_backend_impl.py
+++ b/emu_mps/mps_backend_impl.py
@@ -151,9 +151,7 @@ class MPSBackendImpl:
     def __getstate__(self) -> dict:
         for obs in self.config.observables:
             obs.apply = MethodType(type(obs).apply, obs)  # type: ignore[method-assign]
-        d = {}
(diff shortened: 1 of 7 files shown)


emu_mps/mps_backend_impl.py
```diff
@@ -151,14 +151,16 @@ def __init__(self, mps_config: MPSConfig, pulser_data: PulserData):
     def __getstate__(self) -> dict:
         for obs in self.config.observables:
             obs.apply = MethodType(type(obs).apply, obs)  # type: ignore[method-assign]
-        d = self.__dict__
+        d = {}
+        for key, val in self.__dict__.items():
+            d[key] = val
```
sgrava, line 156: d = self.__dict__.copy()

pablolh (approved):

Issue #88, opened 2025-04-17
fixing _to_abstract_repr


kbidzhiev:
this fails for MPS and SV 
```
    initial_state = MPS.from_state_amplitudes(
        eigenstates=("r", "g"), amplitudes={'rrr': 1.0}
    )
    initial_state._to_abstract_repr()
```

```
    initial_state = StateVector.from_state_amplitudes(
        eigenstates=("r", "g"), amplitudes={'rrr': 1.0}
    )
    initial_state._to_abstract_repr()
```

PR #89, opened 2025-04-17
88 fixing  to abstract repr

kbidzhiev:
fixing #88

### DIFF
--- a/emu_mps/mps.py
+++ b/emu_mps/mps.py
@@ -407,8 +407,8 @@ class MPS(State[complex, torch.Tensor]):
         Returns:
             The resulting MPS representation of the state.s
         """
-        #self._validate_amplitudes(amplitudes, eigenstates)
-        nqubits = len(next(iter(amplitudes.keys())))
+        nqubits = cls._validate_amplitudes(amplitudes, eigenstates)
+        #nqubits = len(next(iter(amplitudes.keys())))
         basis = set(eigenstates)
         if basis == {"r", "g"}:
             one = "r"
--- a/emu_mps/mps.py
+++ b/emu_mps/mps.py
@@ -38,6 +38,7 @@ class MPS(State[complex, torch.Tensor]):
         orthogonality_center: Optional[int] = None,
         config: Optional[MPSConfig] = None,
         num_gpus_to_use: Optional[int] = DEVICE_COUNT,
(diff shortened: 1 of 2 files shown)


pablolh (commented):
Nice
Should the sampled bitstrings from `MPS.sample()` also use the configured eigenstates?
Should emu_sv also use this fix?

kbidzhiev:
@pabloh, honestly, I don't deeply know the reason of introducing other basis options.
depending on basis Operators should track that info as well, op("r","g")*state("0","1") should make sense.

I think fixing this on pulser side would just do the job for all emulators.

docs/emu_mps/notebooks/getting_started.ipynb
```diff
@@ -196,18 +196,15 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
```
pablolh, line 199: do we really need to modify this file?

test/emu_sv/test_state.py
```diff
@@ -123,3 +123,30 @@ def test_from_amplitudes() -> None:
...
+)
+def test_to_abstr_repr(eig, ampl) -> None:
+    initial_state = StateVector.from_state_amplitudes(eigenstates=eig, amplitudes=ampl)
+    abstr = initial_state._to_abstract_repr()
+
+    assert ampl == abstr["amplitudes"]
+    assert eig == abstr["eigenstates"]
+
+
+# @pytest.mark.parametrize(
```
pablolh, line 143: you can keep that one no? check the error message in pytest.raises

docs/emu_mps/notebooks/getting_started.ipynb
kbidzhiev: this `null` happened automatically because I didn't run the file.

I've modified other part. The basis "r,g" was a `set` `{"r", "g"}`. The pulser constructor accepts `Sequence` but not a set

pablolh: ok

test/emu_sv/test_state.py
kbidzhiev: done

pablolh (approved):

Issue #57, opened 2025-03-03
[emu-sv] add custom krylov backward for omega and delta parameters


sgrava:
Make the code differentiable by adding a custom backward to time evolution.

### Actions
 
- In `math`, add a Krylov-based method to compute the derivative of the exponential map $U=e^{-itH}$, with $H(\Omega, \Delta)$. 
- Replace `do_time_step` with a custom `torch.autograd.Function`. Its forward method just computes the time step in the previous way. The backward method returns gradients with respect to $\Omega, \Delta$ and the evolved state.
- document and tests

### Description
It is reasonable to put the core method in math since it is doing a more abstract mathematical operation. In concrete, it is computing the Fréchet [derivative of the exponential map](https://en.wikipedia.org/wiki/Derivative_of_the_exponential_map) as a Gram-Schmidt decomposition, $dU = V^{*}HV$

Support for differentiating through phase and interaction matrix (qubit position) will be added separately later.

PR #90, opened 2025-04-18
double krylov

sgrava:
### Changes
1. Introduce a method `double_krylov` to return a krylov-based decomposition of the derivative of a krylov exp. This piece of software is needed for building the backward of a time step.
2. duplicate code to build a lanczos basis with given tolerance. The alternative would be to change `KrylovExpResults` to allow to store the lanczos basis. But we can do it later after proper testing that
    - all diff workflow works
    - it does not affect performance
Long story short, better safe than sorry.

Related to #57

### Description

Add functionality to compute the derivative of the exponential map along the displacement $|state\rangle\langle grad|$ (for simplicity $|s\rangle\langle g|$).
- Frechet derivative [deifinition](https://nhigham.com/2020/06/23/what-is-a-frechet-derivative/)
- implementation in scipy: [scipy.linalg.expm_frechet](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.expm_frechet.html)

Traditional (and memory expensive) way of doing it is to compute:
```math
\exp
\begin{pmatrix} 
    H & |s\rangle\langle g| \\ 
    0  & H 
\end{pmatrix}
=
\begin{pmatrix} 
    \exp(H) & D_{exp}(H,|s\rangle\langle g| \\ 
    0 & \exp(H) 
\end{pmatrix}
```
Instead, we move the computation on a double Krylov basis
```math
\exp
\begin{pmatrix} 
    H & E \\ 
    0  & H 
\end{pmatrix}
=
\begin{pmatrix} 
    V_s & 0 \\ 
    0 & V_g 
\end{pmatrix}
\exp
\begin{pmatrix} 
    Ts & sg|0\rangle\langle 0|\\ 
    0 & Tg
\end{pmatrix}
\begin{pmatrix} 
    V_s^* & 0\\ 
    0 & V_g^*
\end{pmatrix}
```
```math
=
\begin{pmatrix} 
    V_s & 0 \\ 
    0 & V_g 
\end{pmatrix}
\begin{pmatrix} 
    \exp(Ts) & dS\\ 
    0 & \exp(Tg)
\end{pmatrix}
\begin{pmatrix} 
    V_s^* & 0\\ 
    0 & V_g^*
\end{pmatrix}
```
such that $dU = V_s @ dS @ V_g^*$. With the two Lanczos basis:
```math
V_s = \text{Lanczos}(|s\rangle, H|s\rangle, H^2|s\rangle, \dots)
```
```math
V_g = \text{Lanczos}(|g\rangle, H|g\rangle, H^2|g\rangle, \dots)
```
The derivative is then returned as a decomposition. We don't want to fully compute explicitly $dU$ since it is a matrix $2^N\times 2^N$.

### Benchmark accuracy
We show accuracy in respect to the calculation of the derivative, done in a dense fashion:

```math
\exp
\begin{pmatrix} 
    H & |s\rangle\langle g| \\ 
    0  & H 
\end{pmatrix}
=
\begin{pmatrix} 
    \exp(H) & D_{exp}(H,|s\rangle\langle g|) \\ 
    0 & \exp(H) 
\end{pmatrix}
```
For both methods in consideration `old_double_krylov` and `double_krylov` we compute the norm of the distance $\|dU_{dense}-dU\|$

When the derivative is properly renormalized by the norm of the grad $|grad|$, however, both approaches respect the required tolerance.
The normalization feels natural because the derivative is linearly proportional to the variation $dU \sim |grad|$.

### Alternatives
The derivative we wish to compute can also be cast in the following sum
```math
dU=D_{exp}(H,E)= \sum_{n=0}^{\infty}\sum_{i+j=n}\frac{1}{(i+j)!}H^i|s\rangle\langle g|H^j
```
With the two Lanczos basis
```math
V_s = \text{Lanczos}(|s\rangle, H|s\rangle, H^2|s\rangle, \dots)
```
```math
V_g = \text{Lanczos}(|g\rangle, H|g\rangle, H^2|g\rangle, \dots)
```
a decomposition is readly available as
```math
dU=V_s @ T @ V_g^* 
```
```math
T_{i,j} = \frac{|V_s^{i}||V_g^{j}|}{(i+j)!} 
```


test/emu_base/math/test_double_krylov.py
```diff
@@ -0,0 +1,62 @@
...
+
+    state_block = torch.stack(lanczos_vectors_state)
+    grad_block = torch.stack(lanczos_vectors_grad)
+    dU = state_block.mT @ dS @ grad_block.conj()
+
+    Hsv = dense_rydberg_hamiltonian(omegas, deltas, phis, interactions)
+    E = state.unsqueeze(-1) @ grad.conj().unsqueeze(0)
+    expected_dU = frechet_exp(-1j * dt * Hsv, E)
+
+    assert torch.allclose(dU, expected_dU, atol=tolerance)
```
a-quelle, line 62: put `rtol=0`, it uses as tolerance `atol + val*rtol` and by default `rtol != 0` so you're not testing what you expect, I think.

sgrava: done!

a-quelle (approved):
lgtm

PR #94, opened 2025-04-23
Kb/permute obs

### LINKED ISSUE #39: Permute interaction matrix and output

kbidzhiev:
Permute observables
closes #39


emu_mps/mps_backend.py
```diff
@@ -55,7 +58,15 @@ def run(self) -> Results:
...
         impl.init()  # This is separate from the constructor for testing purposes.
 
-        return self._run(impl)
+        results = self._run(impl)
+        if not self._config.optimize_qubit_ordering:
+            return results
+
+        inv_perm = optimat.inv_permutation(impl.qubit_permutation)
+        permute_bitstrings(inv_perm, results)
+        permute_correlations(inv_perm, results)
```
a-quelle, line 67: also, `Occupation`

emu_mps/mps_backend.py
```diff
@@ -66,3 +77,26 @@ def _run(impl: MPSBackendImpl) -> Results:
...
+        return
+    uuid_bs = results._find_uuid("bitstrings")
+
+    results._results[uuid_bs] = [
+        Counter({optimat.permute_string(bstr, perm): c for bstr, c in bs_counter.items()})
+        for bs_counter in results._results[uuid_bs]
+    ]
+
+
+def permute_correlations(perm: torch.Tensor, results: Results) -> None:
```
a-quelle, line 93: the naming is misleading, a minute ago, I wrote a comment that you forgot to permute the occupation result, yet here it is.

emu_mps/mps_backend_impl.py
```diff
@@ -118,8 +120,17 @@ def __init__(self, mps_config: MPSConfig, pulser_data: PulserData):
         self.timestep_count: int = self.omega.shape[0]
         self.has_lindblad_noise = pulser_data.has_lindblad_noise
         self.lindblad_noise = torch.zeros(2, 2, dtype=torch.complex128)
-        self.full_interaction_matrix = pulser_data.full_interaction_matrix
-        self.masked_interaction_matrix = pulser_data.masked_interaction_matrix
+        self.qubit_permutation = (
+            optimat.minimize_bandwidth(pulser_data.full_interaction_matrix)
```
a-quelle, line 124: please also add this to the qubit_order field in the `Results`. Currently, it's left empty, but this is really an error on our side.

emu_mps/mps_backend.py
a-quelle, line 67: please also update the qubit_order in the results again to reflect the original ordering. Maybe split this logic into a new function? Something like `reorder_results`, and then you can just do
`return self.reoder_results(results)`, where it handles the lines 62 and below.

emu_mps/mps_backend_impl.py
```diff
@@ -203,6 +214,18 @@ def init_initial_state(self, initial_state: State | None = None) -> None:
             )
 
         assert isinstance(initial_state, MPS)
+        if not torch.equal(
+            self.qubit_permutation, optimat.eye_permutation(self.qubit_count)
+        ):
+            # permute the initial state to match with permuted Hamiltonian
+            abstr_repr = initial_state._to_abstract_repr()
```
a-quelle, line 221: this will fail if the initial state was not created via `MPS.from_state_amplitudes`, but for example as output from a `StateResult` observable. On a similar note, you don't reorder the state in a `StateResult` when reordering the results (for obvious reasons). I think we should add some logic that when optimatrix is used, `StateResult` is not allowed as an observable, and that if `initial_state._to_abstract_repr()` we add a more graceful error that state reordering is only supported when the input state is constructed from abstract representation.

test/emu_mps/test_endtoend.py
```diff
@@ -217,7 +221,82 @@ def test_XY_3atomswith_slm() -> None:
...
+    bitstrings = result.bitstrings[ntime_step]
+    occupation = result.occupation[ntime_step]
+    cor_mat = result.correlation_matrix[ntime_step]
+    energy = result.energy[ntime_step]
+    energy_variance = result.energy_variance[ntime_step]
+
+    expect_occup = torch.tensor([1, 1, 1, 0, 0, 0], dtype=torch.float64)
+    expect_corr = torch.outer(expect_occup, expect_occup).to(torch.complex128)
+
+    assert bitstrings["111000"] == 100
```
a-quelle, line 282: since I asked you to also fill in the `qubit_order` in `Results` perhaps also add an assert on the order in this test.

test/emu_mps/test_endtoend.py
```diff
@@ -217,7 +221,82 @@ def test_XY_3atomswith_slm() -> None:
...
+        True,
+    ],
+)
+def test_end_to_end_domain_wall_ring(
+    optimize_order: bool,
+) -> None:
+    torch.manual_seed(seed)
+
+    num_qubits = 6
+    seq = pulser_afm_sequence_ring(
```
a-quelle, line 237: what's the meaning of this sequence? If I interpret it correctly, you start from a "rrrggg" state, and then have omega=0, so the state only acquires a phase as time goes on, which is the time integral of `H[56,56]` over time, considering `H` as a dense matrix. I understand that it's a nice system you can solve analytically, but what does it add to the existing tests?

test/emu_mps/test_endtoend.py
```diff
@@ -231,33 +310,36 @@ def test_end_to_end_afm_ring() -> None:
...
+    assert approx(occupation, rel=1e-3) == [0.5782] * 10
 
-    energy_variance = result.energy_variance[-1]
-    assert approx(energy_variance, 1e-6) == 45.905980469959104
+    assert approx(energy, abs=1e-2) == -115.34554479213088
+    assert approx(energy, rel=1e-4) == -115.34554479213088
 
-    second_moment_energy = result.energy_second_moment[-1]
-    assert approx(second_moment_energy, 1e-6) == 13350.07680148
+    assert approx(energy_variance, abs=1e-1) == 45.91111056399
```
pablolh, line 338: mmh abs=1e-1 and 45.91111056399
`45.9` would be good

test/emu_mps/test_endtoend.py
```diff
@@ -60,6 +60,9 @@ def simulate(
     initial_state: Any | None = None,
     given_fidelity_state: bool = True,
     interaction_cutoff: float = 0,
+    eval_times: list[float] = [1.0],
+    n_shots: int = 1000,
```
pablolh, line 64: you don't use `n_shots` as parameter

test/emu_mps/test_endtoend.py
pablolh: Instead of modifying this existing test by changing tolernaces, what about creating (a) new test(s) that checks that observables are identical/similar between with and without optimatrix? That lets us also see what is the difference to expect. So simulate twice same sequence with and without optimatrix

emu_mps/mps_config.py
```diff
@@ -84,6 +85,7 @@ def __init__(
             max_krylov_dim=max_krylov_dim,
             extra_krylov_tolerance=extra_krylov_tolerance,
             num_gpus_to_use=num_gpus_to_use,
+            optimize_qubit_ordering=optimize_qubit_ordering,
```
pablolh, line 88: is that also an idea to let users specify their own qubit ordering?
If someone simulates same sequence multiple times with e.g. different dt, they shouldn't have to run optimatrix more than once.
If someone wants to use a different solver (Cplex), that would also be possible

emu_mps/mps_backend.py
kbidzhiev: My point was that it's a one- and two-point correlation functions. I'll give a better name then

test/emu_mps/test_endtoend.py
kbidzhiev: previous tests were benchmarking emu-mps against emu-mps itself. We take a run and consider its results as correct. Now I do it against emu-sv. I compare here permuted and unpermuted emu-mps against emu-sv and show that quality didn't drop with permutation.

pablolh: is it an idea to use emu-sv directly instead of hardcoding values like `45.91111056399`? With new API it should be really easy to implement

emu_mps/mps_config.py
kbidzhiev: User can specify their order initially when preparing an interaction matrix according to their needs. So no use of the optimatrix is required. 

I would say that this trick with permuting indices is new, and not known in the community. I used Cplex and it takes several minutes to permute the matrix, while optimatrix does basic heuristics quickly around a second. This PR is a first order improvement, - adding a new optimisation solver can be considered as a next step

emu_mps/mps_backend_impl.py
kbidzhiev: shame on me, didn't notice

test/emu_mps/test_endtoend.py
a-quelle: ok, good.

emu_mps/mps_backend.py
```diff
@@ -55,7 +58,15 @@ def run(self) -> Results:
         impl = create_impl(self._sequence, self._config)
         impl.init()  # This is separate from the constructor for testing purposes.
 
-        return self._run(impl)
+        results = self._run(impl)
+        if not self._config.optimize_qubit_ordering:
```
pablolh, line 62: I would extract this to separate function, something like
```
results = self._run(impl)

return reorder(results, config)
```

emu_mps/optimatrix/permutations.py
```diff
@@ -1,6 +1,33 @@
 import torch
 
 
+def eye_permutation(n: int) -> torch.Tensor:
```
pablolh, line 4: So, I have an idea for a future refactor:
create a `Permutation` class with methods
```
class Permutation:
  def __init__(self, ...):
    # ...


  @classmethod
  def id(cls, n: int):
    return cls(torch.arange(n))

 
  def is_identity(self):
    return torch.equals(self.perm, torch.arange(len(self.perm)))


  def apply(self, input: list | str | torch.Tensor):
    if isinstance(input, list):
      return [input_list[i] for i in self.perm.tolist()]
    elif isinstance(input, str):
      return "".join(self.apply(list(input)))
    elif isinstance(input, torch.Tensor):
      # ...


  def inverse(self) -> Permutation:
    inv_perm = torch.empty_like(self.perm)
    inv_perm[self.perm] = np.arange(len(self.perm))
    return Permutation(inv_perm) 
```

emu_mps/mps_backend_impl.py
kbidzhiev: done

emu_mps/mps_backend.py
kbidzhiev: done

emu_mps/mps_backend_impl.py
kbidzhiev: now `MPSconfig` checks observables and raises an error if the permutation if not supported

test/emu_mps/test_endtoend.py
kbidzhiev: done in the other test 'domain_wall'

test/emu_mps/test_endtoend.py
kbidzhiev: I'm a bit hesitant here. The idea is good, I'm afraid it implementation will require changes in other places. Propose to do that separately. For me it sounds like uniformising and generalising the tests

test/emu_mps/test_endtoend.py
kbidzhiev: removed

emu_mps/mps_backend.py
kbidzhiev: now its `impl.permute_results(results)`

emu_mps/optimatrix/permutations.py
kbidzhiev: why not !

emu_mps/mps_backend_impl.py
```diff
@@ -510,6 +540,45 @@ def fill_results(self) -> None:
...
+            self.qubit_permutation, optimat.eye_permutation(self.qubit_count)
+        ):
+            inv_perm = optimat.inv_permutation(self.qubit_permutation)
+            permute_bitstrings(results, inv_perm)
+            permute_occupations_and_correlations(results, inv_perm)
+            permute_atom_order(results, inv_perm)
+        return results
+
+
+def permute_bitstrings(results: Results, perm: torch.Tensor) -> None:
```
pablolh, line 554: those need to be unit-tested

pablolh: the more small, short, good unit tests you have, the less messy, complicated integration tests you have to write

emu_mps/mps_backend_impl.py
```diff
@@ -128,7 +141,12 @@ def __init__(self, mps_config: MPSConfig, pulser_data: PulserData):
         self.swipe_direction = SwipeDirection.LEFT_TO_RIGHT
         self.tdvp_index = 0
         self.timestep_index = 0
-        self.results = Results(atom_order=(), total_duration=self.target_times[-1])
+        self.results = Results(
+            atom_order=tuple(
+                optimat.permute_list(list(pulser_data.qubit_ids), self.qubit_permutation)
```
a-quelle, line 146: I'm not sure if it's a good idea to store the permutation too. This field is part of the general API definition, and it's supposed to be a tuple of strings, i.e. just the `qubit_id`s in the permuted order.

kbidzhiev: mypy changes a bit readability. The code essentially is

```
        old_order = list(pulser_data.qubit_ids)  # it's stored as a tuple
        permuted_order = optimat.permute_list(old_order, self.qubit_permutation)
        self.results = Results(
            atom_order=tuple(permuted_order),
            total_duration=self.target_times[-1],
        )
```

I store in `Results` only the permuted `qubit_ids`.

Did you mean smth else ?

emu_mps/mps_backend_impl.py
kbidzhiev: done

pablolh (approved):

emu_mps/mps_backend_impl.py
a-quelle: yes, you store in atom_order an object of type `tuple[list, list]` while it should be a tuple of type `tuple[str,...]` specifically, it should contain the first item from what you assign, cast to a tuple.

kbidzhiev: have a look

Issue #93, opened 2025-04-22
[Emu-SV]: Noise - Master Equation: Implementation of the lindblad operator action to a density matrix


murogrande:
**Noise implementation by solving the Lindblad master equation**

The Linbdlad equation is given by 
$\frac{d\rho}{dt} = -i [H,\rho(t)] + \sum_k A_k\rho(t)A_k^ \dagger-\frac{1}{2}A_k^\dagger A_k \rho(t) -\frac{1}{2}\rho(t)A_k^\dagger A_k$. 

Create a class where the action of the lindblad operator can be applied to a matrix (if it is working in the case of a generic matrix it can work in the case of the density matrix). The implementation should not contain kronecker products. 

- Add tests 

The implementation of the code has been tested against Qutip `mesolver`. In the case of emu-sv, the evolution was done using `krylov_exp` implementation. The results are for a first step with `dt=10` in the evolution of a quantum system of 2 particles with 3 lindblad operators. 
The initial state is given by  
 - $\rho = |11><11|$ 
 - and the Lindblad operators are the following 
$L_1 = \sqrt{\frac{1}{3}} \sigma^-$, $L_2 = \sqrt{\frac{1}{3}} \sigma_x$, $L_3 = \sqrt{\frac{1}{3}} \sigma_y$

 Post the screenshots as evidence. On the le

<img src="https://github.com/user-attachments/assets/f7a405b0-6484-4b0d-96e4-e6c3cd93529a" width="49%">
<img src="https://github.com/user-attachments/assets/e8a79fd2-2f9e-4e25-a74f-4f95ab95afd4" width="49%">

PR #96, opened 2025-04-24
[emu-sv] noise: lindblad operator implementation

murogrande:
Related to #93


emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,136 @@
...
+            diag = diag.view(2**i, 2, -1)
+            i_fixed = diag[:, 1, :]
+            for j in range(i + 1, self.nqubits):
+                i_fixed = i_fixed.view(2**i, 2 ** (j - i - 1), 2, -1)
+                # replacing i_j_fixed by i_fixed breaks the code :)
+                i_j_fixed = i_fixed[:, :, 1, :]
+                i_j_fixed += self.interaction_matrix[i, j]
+        return diag.view(-1)
+
+    def apply_local_operator_to_density_matrix_to_local_op(
```
pablolh, line 49: `apply_local_operator_to_density_matrix_to_local_op` is a weird name no?

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        nqubits = self.nqubits
+        dim = 2 ** (nqubits - 1)
+
+        # reshape density matrix to 2n-dimensional tensor
+        rho = density_matrix.view([2] * 2 * nqubits)
+
+        # permute to bring target_qubit to front in both bra and ket indices
+        perm = list(range(nqubits))
+        perm.remove(target_qubit)
```
pablolh, line 69: ~what about something like~
```
r = nqubits - target_qubit - 1
rho = rho.view(2 ** target_qubit, 2, 2 ** r, 2 ** target_qubit, 2, 2**r)
rho = rho.permute(1, 4, 0, 2, 3, 5)
```


EDIT: I think it's quite costly to permute and then call `contiguous` on a big tensor like this.

What about using `tensordot` and specifying the dimensions to contract instead of doing matrix multiplication? I think it will make the code shorter and more efficient. In fact, no call to `permute` is required, since you permute back at the end.

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,136 @@
...
+                omega * sigmax.to(device=self.device)
+                - delta * n_op.to(device=self.device)
+                + sum_lindblad_local
+            )
+            H_local_rho += self.apply_local_operator_to_density_matrix_to_local_op(
+                density_matrix, H_q, qubit
+            )
+
+        # apply diagonal interaction  ∑ᵢⱼ Uᵢⱼ nᵢ nⱼ
+        diag_term = self.diag.view(-1, 1) * density_matrix  # elementwise column scaling
```
pablolh, line 123: should we always call the density matrix `rho`? it's quite confusing to me

emu_sv/lindblad_operator.py
pablolh: Concerning the function itself, I think personally I would enjoy more a pair of functions `apply_local_op_left` and `apply_local_op_right`

emu_sv/lindblad_operator.py
a-quelle: In physics calling the density matrix `rho` is kind of the same as calling a variable `x` in calculus, meaning it's kind of standard, but it's ok to deviate for the sake of clarity.

emu_sv/__init__.py
```diff
@@ -15,6 +15,7 @@
 from .dense_operator import DenseOperator
 from .sv_backend import SVBackend, SVConfig
 from .state_vector import StateVector, inner
+from .lindblad_operator import LindbladOperator
```
a-quelle, line 18: Do we expect the user to interface with this class directly? If so, why?

emu_sv/sv_backend.py
```diff
@@ -35,6 +35,8 @@ def run(self) -> Results:
             sequence=self._sequence, config=self._config, dt=self._config.dt
         )
         self.target_times = pulser_data.target_times
+        # list_lindblad_op = pulser_data.lindblad_ops  # reading lindblad operators
```
a-quelle, line 38: why is this commented out?

murogrande: This is me thinking how to extract the Lindblad operators. Sorry, I should have deleted before.
This will be implemented when I update the API

emu_sv/__init__.py
murogrande: This is for testing purposes, I will delete this

emu_sv/lindblad_operator.py
murogrande: We should unified name,  I will change to `density_matrix`.

emu_sv/lindblad_operator.py
murogrande: please check how we improved this part of the code with @a-quelle

emu_sv/lindblad_operator.py
murogrande: do you like just `apply_local_op_to_density_matrix`?

emu_sv/time_evolution.py
```diff
@@ -21,6 +23,18 @@ def do_time_step(
         device=state_vector.device,
     )
     op = lambda x: -1j * dt * (ham * x)
+
+    def op_noise(x: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
murogrande: Just me thinking how the code will look like. Sorry, I will delete this

emu_sv/lindblad_operator.py
murogrande: How much are you against of having 1 function? 
Is something that really bothers you?

emu_sv/lindblad_operator.py
a-quelle: that seems like a good name

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,118 @@
...
+            self.device
+        )
+
+        # apply local Hamiltonian terms (Ω σₓ - δ n - 0.5i ∑ₖ Lₖ† Lₖ) to each qubit
+        H_local_den_matrix = torch.zeros_like(
+            density_matrix, dtype=dtype, device=self.device
+        )
+        for qubit, (omega, delta) in enumerate(zip(self.omegas, self.deltas)):
+            H_q = (
+                omega * sigmax.to(device=self.device)
```
a-quelle, line 94: what about phi?

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,118 @@
...
+        if op_conj_T:
+            density_matrix = density_matrix.view(
+                2 ** (target_qubit + self.nqubits), 2, -1
+            )
+            density_matrix = local_op.conj() @ density_matrix
+            density_matrix = density_matrix.view(orignal_shape)
+
+        return density_matrix
+
+    def __matmul__(self, density_matrix: torch.Tensor) -> torch.Tensor:
```
a-quelle, line 76: you could probably reduce memory consumption a bit by reusing variables. For example after line 107, you have in memory `H_local_den_matrix` `H_den_matrix` and `commutator`, while only the last one is still needed at that point.

emu_sv/lindblad_operator.py
pablolh: concerning the contiguous calls, I discovered yesterday that it might not be as easy as I thought. I'll look at the new version, and as you said correctness comes before performance.

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,118 @@
+import torch
+from emu_base.lindblad_operators import compute_noise_from_lindbladians
+
+
+dtype = torch.complex128
+sigmax = torch.tensor([[0.0, 1.0], [1.0, 0.0]], dtype=dtype)
+n_op = torch.tensor([[0.0, 0.0], [0.0, 1.0]], dtype=dtype)
+
+
+class LindbladOperator:
```
pablolh, line 10: Is this really what a lindblad operator is?
I already called lindblad operator the individual noise jump operators, what is correct terminology?
For me, this class is the noisy hamiltonian for density matrices, but with a different equation...

pablolh:
new version with `view` looks good! So the real problem with `tensordot` using extra memory is when you contract multiple non-consecutive dimensions that are not at the beginning or end, I think.

emu_sv/lindblad_operator.py
murogrande: you are right, let me add it. Thanks

murogrande: I added the complex part, also I update the tests in order to support this

emu_sv/lindblad_operator.py
murogrande: change folder name `lindblad_operator` -> `jump_limblad_operators`

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,159 @@
...
+        Calculate a local operator (2x2) L being multiplied by a density matrix ρ
+        from the left
+        Return L @ ρ
+        """
+
+        orignal_shape = density_matrix.shape
+        density_matrix = density_matrix.view(2**target_qubit, 2, -1)
+        density_matrix = local_op @ density_matrix
+
+        # if op_conj_T:
```
pablolh, line 67: remove commented part?

emu_sv/lindblad_operator.py
sgrava: I understand that there is confusion, see [here](https://en.wikipedia.org/wiki/Lindbladian).
You can call this guy either Lindbladian or Liouvillian

test/emu_sv/test_linblad_operator.py
```diff
@@ -0,0 +1,191 @@
...
+    listdag = []
+    for i in range(nqubits):
+        if i == target_qubit:
+            listdag.append(lindblad_op.conj().T.contiguous())
+        else:
+            listdag.append(ident)
+
+    resa = torch.kron(lista[0], lista[1])
+    resdag = torch.kron(listdag[0], listdag[1])
+    for i in range(2, nqubits):
```
pablolh, line 111: Please write a function to compute those kron products of lists, if it doesn't exist yet
let's factorize code and make it easier to read

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,118 @@
...
+                i_j_fixed = i_fixed[:, :, 1, :]
+                i_j_fixed += self.interaction_matrix[i, j]
+        return diag.view(-1)
+
+    def apply_local_op_to_density_matrix(
+        self,
+        density_matrix: torch.Tensor,
+        local_op: torch.Tensor,
+        target_qubit: int,
+        op_conj_T: bool = False,
```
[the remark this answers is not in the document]
murogrande: I changed to `apply_local_op_to_density_matrix` in the case of $L \rho$ and `apply_density_matrix_to_local_op_T` in the case of $\rho L^\dagger$

emu_sv/lindblad_operator.py
murogrande: done

test/emu_sv/test_linblad_operator.py
murogrande: I added a `list_2_kron` function and save it in `utils_testing`

sgrava: consider `functools.reduce(torch.kron, list)`

emu_sv/lindblad_operator.py
```diff
@@ -0,0 +1,189 @@
...
+                    )
+                    - delta * n_op.to(device=self.device)
+                    + sum_lindblad_local
+                )
+                H_den_matrix += self.apply_local_op_to_density_matrix(
+                    density_matrix, H_q, qubit
+                )
+
+        # apply the interaction terms  ∑ᵢⱼ Uᵢⱼ nᵢ nⱼ
+        diag_term = self.diag.view(-1, 1) * density_matrix
```
a-quelle, line 170: `H_den_matrix += RHS`. Currently, `diag_term` will linger in memory until the end of the function.

murogrande: sure !

test/emu_sv/test_linblad_operator.py
murogrande: nice point !

emu_base/jump_lindblad_operators.py
```diff
@@ -45,3 +45,22 @@ def get_lindblad_operators(
...
+        lindbladian.shape == (2, 2) for lindbladian in lindbladians
+    ), "Only single-qubit lindblad operators are supported"
+
+    return (
+        -1j
+        / 2.0
+        * sum(
+            (lindbladian.T.conj() @ lindbladian for lindbladian in lindbladians),
+            start=torch.zeros(2, 2, dtype=torch.complex128),
+        )
```
sgrava, line 65:
```suggestion
    return (
        -1j
        / 2.0
        * sum(
            (lindbladian.T.conj() @ lindbladian for lindbladian in lindbladians),
            start=torch.zeros(2, 2, dtype=torch.complex128),
        )
```
```
return -0.5j*sum(L.mH @ L for L in lindbladians)
```

emu_sv/time_evolution.py
```diff
@@ -12,6 +13,7 @@ def do_time_step(
     full_interaction_matrix: torch.Tensor,
     state_vector: torch.Tensor,
     krylov_tolerance: float,
+    linblad_ops: Optional[list[torch.Tensor]],
```
pablolh, line 16: why pass this parameter if you are not using it?

pablolh: also, if you type it as Optional, then you shouldn't have to modify existing invocations like you did below

murogrande: This parameter is for the next steps, do you want to delete it?  But I will add this in the future.

pablolh: There is no reason to add it to this change.
What if you or we decide to do differently?

murogrande: ok, I will delete this

pablolh (approved):

PR #97, opened 2025-04-24
Sg/do time step as autograd function

sgrava:
Support gradient backpropagation for time steps where $\Omega$, $\Delta$ or $\Phi$ require it.

### Changes
- `do_time_step` is now a `torch.autograd.Function`
    - forward simply do a krylov time step as before
    - backward propagate the gradient of the output state to the input (as prescribed in [torch docs](https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html))
- `backward` is tested against dense linear algebra, results within imposed relative tolerance

### Benchmark
 [new run](https://gitlab.pasqal.com/emulation/rydberg-atoms/benchmarks_github_gitlab/-/jobs/1467498/artifacts/file/benchmarks/emu_sv_cpu_gpu/results/emu_sv_runtimes.png) vs [old run](https://gitlab.pasqal.com/emulation/rydberg-atoms/benchmarks_github_gitlab/-/jobs/1463051/artifacts/file/benchmarks/emu_sv_cpu_gpu/results/emu_sv_runtimes.png) of the `emu_sv_cpu_gpu` benchmark show same runtime and memory consumption in the logs below.

For 20 qubits - GPU
new
```
step = 448/450, RSS = 260.182 MB, Δt = 0.014 s
step = 449/450, RSS = 260.182 MB, Δt = 0.014 s
step = 450/450, RSS = 260.182 MB, Δt = 0.077 s
```
old
```
step = 448/450, RSS = 260.182 MB, Δt = 0.014 s
step = 449/450, RSS = 260.182 MB, Δt = 0.014 s
step = 450/450, RSS = 260.182 MB, Δt = 0.074 s
```

For 20 qubits - CPU
new
```
step = 448/450, RSS = 1326.564 MB, Δt = 0.189 s
step = 449/450, RSS = 1326.564 MB, Δt = 0.157 s
step = 450/450, RSS = 1326.564 MB, Δt = 0.191 s
```
old
```
step = 448/450, RSS = 1300.072 MB, Δt = 0.195 s
step = 449/450, RSS = 1300.072 MB, Δt = 0.201 s
step = 450/450, RSS = 1300.072 MB, Δt = 0.211 s
```

### Minor changes
- small fix in `test_dense_operator.py ` to get rid of a warning in tests

### Disscusion
- testing also with `gradcheck`
I propose I open a separate issue for it

### DIFF
--- a/emu_sv/time_evolution.py
+++ b/emu_sv/time_evolution.py
@@ -7,26 +7,28 @@ from emu_sv.hamiltonian import RydbergHamiltonian
 
 def _apply_omega_real(
     result: torch.Tensor,
-    dim: int,
+    i: int,
     inds: torch.Tensor,
     source: torch.Tensor,
     alpha: complex,
 ) -> None:
-    result.index_add_(dim, inds, source, alpha=alpha)
+    """Accumulate to `result` the application of ασˣᵢ on `source`"""
+    result.index_add_(i, inds, source, alpha=alpha)
 
 
 def _apply_omega_complex(
     result: torch.Tensor,
-    dim: int,
+    i: int,
     inds: torch.Tensor,
     source: torch.Tensor,
     alpha: complex,
 ) -> None:
-    result.index_add_(dim, inds[0], source[:, :, 0, :].unsqueeze(2), alpha=alpha)
+    """Accumulate to `result` the application of ασ⁺ᵢ + α*σ⁻ᵢ on `source`"""
+    result.index_add_(i, inds[0], source.select(i, 0).unsqueeze(i), alpha=alpha)
     result.index_add_(
-        dim,
+        i,
         inds[1],
-        source[:, :, 1, :].unsqueeze(2),
+        source.select(i, 1).unsqueeze(2),
         alpha=alpha.conjugate(),
     )
 
@@ -42,7 +44,7 @@ class DHDOmegaSparse:
     def __init__(self, index: int, device: str, nqubits: int, phi: torch.Tensor):
         self.index = index
         self.shape = (2**index, 2, 2 ** (nqubits - index - 1))
-        self.inds = torch.tensor([1, 0], device=device)
+        self.inds = torch.tensor([1, 0], device=device)  # flips the state, for 𝜎ₓ
         self.alpha = 0.5 * torch.exp(1j * phi).item()
         if phi != 0:
             self._apply_sigmas = _apply_omega_complex
@@ -78,13 +80,7 @@ class DHDPhiSparse:
     def __matmul__(self, vec: torch.Tensor) -> torch.Tensor:
         vec = vec.view(vec.shape[0], *self.shape)  # add batch dimension
         result = torch.zeros_like(vec)
-        result.index_add_(2, self.inds[0], vec[:, :, 0, :].unsqueeze(2), alpha=self.alpha)
-        result.index_add_(
-            2,
-            self.inds[1],
-            vec[:, :, 1, :].unsqueeze(2),
-            alpha=self.alpha.conjugate(),
-        )
+        _apply_omega_complex(result, 2, self.inds, vec, alpha=self.alpha)
         return result.view(vec.shape[0], -1)
 
 
--- a/emu_sv/time_evolution.py
+++ b/emu_sv/time_evolution.py
@@ -9,6 +9,8 @@ class DHDOmegaSparse:
     """
     Derivative of the RydbergHamiltonian respect to Omega.
         ∂H/∂Ωₖ = 0.5[cos(ϕₖ)σˣₖ + sin(ϕₖ)σʸₖ]
+
+    If ϕₖ=0, simplifies to ∂H/∂Ωₖ = 0.5σˣₖ
     """
 
     def __init__(self, index: int, device: str, nqubits: int, phi: torch.Tensor):
@@ -18,7 +20,7 @@ class DHDOmegaSparse:
         self.alpha = 0.5 * torch.exp(1j * phi).item()
         if phi != 0:
             self._apply_sigmas = self._apply_omega_complex
-        else:
+        else:  # ∂H/∂Ωₖ = 0.5σˣₖ
             self._apply_sigmas = self._apply_omega_real
 
     def __matmul__(self, vec: torch.Tensor) -> torch.Tensor:
@@ -247,9 +249,7 @@ class EvolveStateVector(torch.autograd.Function):
                 dho = DHDOmegaSparse(i, e_l.device, nqubits, phis[i])
                 # compute the trace
                 v = dho @ e_l
-                grad_omegas[i] = (
-                    -1j * dt * torch.tensordot(Vg.conj(), v, dims=([0, 1], [0, 1]))
-                ).real
+                grad_omegas[i] = (-1j * dt * torch.tensordot(Vg.conj(), v)).real
 
         if ctx.needs_input_grad[2]:
             grad_deltas = torch.zeros_like(deltas)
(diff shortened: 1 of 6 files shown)


emu_sv/time_evolution.py
```diff
@@ -1,32 +1,206 @@
...
+            device=state.device,
+        )
+
+        if any(ctx.needs_input_grad[1:4]):
+            op = lambda x: -1j * dt * (ham * x)
+            lanczos_vectors_state, dS, lanczos_vectors_grad = double_krylov(
+                op, state, grad_state_out, tolerance
+            )
+            Vs = torch.stack(lanczos_vectors_state)
+            Vg = torch.stack(lanczos_vectors_grad)
```
a-quelle, line 176: you can del the two lists of lanczos vectors. It will save memory, since the stacked tensor does not reuse the storage, and new memory is allocated.

sgrava: done, please, have a look!

emu_sv/time_evolution.py
```diff
@@ -1,32 +1,274 @@
...
+
+    If ϕₖ=0, simplifies to ∂H/∂Ωₖ = 0.5σˣₖ
+    """
+
+    def __init__(self, index: int, device: str, nqubits: int, phi: torch.Tensor):
+        self.index = index
+        self.shape = (2**index, 2, 2 ** (nqubits - index - 1))
+        self.inds = torch.tensor([1, 0], device=device)
+        self.alpha = 0.5 * torch.exp(1j * phi).item()
+        if phi != 0:
```
a-quelle, line 21: we just discovered last week that this pattern is toxic. Let's refactor

pablolh: I think the function assignment is good now cuz you are using static/global functions

I think better use something like `torch.is_nonzero` rather than `!= 0`

emu_sv/time_evolution.py
```diff
@@ -1,32 +1,281 @@
...
-            state_vector,
+            state,
             norm_tolerance=krylov_tolerance,
             exp_tolerance=krylov_tolerance,
-        ),
-        ham,
-    )
+            is_hermitian=True,
+        )
+        ctx.save_for_backward(omegas, deltas, phis, interaction_matrix, state)
```
pablolh, line 154: should that be conditional to the fact that you are requiring differentiability on tensors?
Or is that automatic in Torch?

pablolh: (make sure this change doesn't impact performance of the normal simulations)

test/emu_sv/test_time_evolution.py
```diff
@@ -54,10 +54,118 @@ def test_forward_with_phase(N: int, krylov_tolerance: float) -> None:
...
+    interactions = randn_interaction_matrix(N)
+
+    ham_params = (omegas, deltas, phis, interactions)
+
+    state = torch.randn(2**N, dtype=dtype, requires_grad=True)
+    state = state / state.norm()
+
+    dt = 1.0  # big timestep 1 μs
+
+    Op = torch.randn(2**N, 2**N, dtype=dtype)
```
pablolh, line 122: no capital letters in variable names, it messes in particular the syntax coloring in github

test/emu_sv/test_time_evolution.py
```diff
@@ -54,10 +54,118 @@ def test_forward_with_phase(N: int, krylov_tolerance: float) -> None:
...
+
+
+dtype_params = torch.float64
+
+
+@pytest.mark.parametrize(
+    "N, krylov_tolerance",
+    [(n, tol) for n in [5, 8] for tol in [1e-8, 1e-10, 1e-12]],
+)
+def test_backward(N, krylov_tolerance):
```
pablolh, line 73: I am not sure this test is useful, since `test_backward_Op` is more generic and the difference would just be torch function applications, you don't need to test the grads of that, plus I see a lot of duplication

test/emu_sv/test_time_evolution.py
```diff
@@ -54,10 +54,118 @@ def test_forward_with_phase(N: int, krylov_tolerance: float) -> None:
...
+
+    for grad, ed_grad in zip(grads, ed_grads):
+        assert torch.allclose(grad, ed_grad, rtol=krylov_tolerance)
+
+
+@pytest.mark.parametrize(
+    "N, krylov_tolerance",
+    [(n, tol) for n in [4, 6] for tol in [1e-8, 1e-12]],
+)
+def test_backward_with_phase(N, krylov_tolerance):
```
pablolh, line 143: you are correct in your comment about fixtures as a way to factorize code for ham parameters, code duplication is bad for multiple reasons, one of them is clarity for whoever is reading the code.
For instance, it took me way too long to see that the only difference in this test is the tuple parameter to `torch.autograd.grad`.

BTW I think the grads are computed independently, and as such, I am not sure that making a different test for adding phase as computed grad is useful - just compute it in the second test

emu_sv/time_evolution.py
```diff
@@ -1,32 +1,281 @@
...
+        return result.view(vec.shape[0], -1)
+
+
+class DHDDeltaSparse:
+    """
+    Derivative of the Rydberg Hamiltonian respect to Delta:
+        ∂H/∂Δₖ = -nₖ
+    """
+
+    def __init__(self, index: int, device: str, nqubits: int):
```
murogrande, line 93: `device: str` -> `device:torch.device`

test/emu_sv/test_time_evolution.py
sgrava: removed in refactoring tests. See below.

test/emu_sv/test_time_evolution.py
sgrava: I agree! I refactored the tests.
Plz, have a look!

test/emu_sv/test_time_evolution.py
sgrava: > BTW I think the grads are computed independently, and as such, I am not sure that making a different test for adding phase as computed grad is useful - just compute it in the second test

I agree, removed phase/no phase in tests

test/emu_sv/test_time_evolution.py
```diff
@@ -2,57 +2,100 @@
...
     dense_rydberg_hamiltonian,
-    nn_interaction_matrix,
     randn_interaction_matrix,
 )
-from emu_sv.time_evolution import do_time_step
+from emu_sv.time_evolution import EvolveStateVector
 
 dtype = torch.complex128
+dtype_params = torch.float64
 device = "cpu"
```
murogrande, line 11: try with gpu, all the backward tests are failing

murogrande (commented):
it is fine for me.

test/emu_sv/test_time_evolution.py
sgrava: Ah good catch! I'll fix it

test/emu_sv/test_time_evolution.py
pablolh: nice

test/emu_sv/test_time_evolution.py
sgrava: Uhhh nice!
Basically, the T matrix in `krylov_exp` is always on CPU.
Normally this is fine since we don't need it as a tensor, but in `double_krylov` yes.

emu_sv/time_evolution.py
sgrava: Done! take a look at the benchmark above.
Looks good to me.

test/emu_sv/test_time_evolution.py
sgrava: Done!
- `test_double_krylov.py`
-` test_time_evolution.py`
also works on GPU

pablolh (approved):

PR #99, opened 2025-04-28
small memory optimization

a-quelle:
While we're at it.
This saves a single state vector of memory in each iteration after the first.

### DIFF
--- a/emu_sv/sv_backend.py
+++ b/emu_sv/sv_backend.py
@@ -90,6 +90,7 @@ class SVBackend(EmulatorBackend):
                 self.results,
             )
             self.time = time.time()
+            del H
 
         return self.results
 


pablolh:
What is the exact reason because it saves memory?
My idea is that "do_time_step" is computed (and therefore allocates memory) before the H variable is "erased" and the previous H contents deallocated, therefore for a small time you have two "H" in memory. Am I correct?

a-quelle:
> What is the exact reason why it saves memory? My idea is that "do_time_step" is computed (and therefore allocates memory) before the H variable is "erased" and the previous H contents deallocated, therefore without this `del` for a small time you have two "H" in memory. Am I correct?

Indeed, and this "small time" is precisely when the krylov algorithm runs and peak memory usage happens.

PR #101, opened 2025-04-29
Ground state solver

elmerhej:
Milestone 1: in DMRG project: writing a Lanczos algorithm for energy minimization using Krylov truncated subspace

PR includes:
-  the Lanczos algo script
-  unit tests


murogrande:
Be careful with referencing your issues. If you click in Milestone `#1` , will bring to the first issue in the repo.

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+        ground_energy = eigvals[0].real
+        ground_eigenvector = eigvecs[:, 0]
+        prev_eigen_vec = ground_eigenvector.clone()
+
+        # Residual convergence check
+        residual_norm = torch.norm(
+            T_truncated @ ground_eigenvector - ground_energy * ground_eigenvector
+        )
+        if residual_norm < residual_tolerance:
+            converged = True
```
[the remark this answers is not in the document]
a-quelle: In any case, I think it'd be good to have a test that checks you converge in the expected number of iterations.

test/emu_base/math/test_krylov_energy_min.py
```diff
@@ -0,0 +1,104 @@
...
+    result: KrylovEnergyResult = krylov_energy_minimization_impl(
+        op=op,
+        v=v,
+        norm_tolerance=1e-12,
+        residual_tolerance=1e-6,
+        max_krylov_dim=dim,
+    )
+
+    assert result.converged is True
+    assert result.happy_breakdown is True
```
pablolh, line 35: you need a test where `converged` is `True` but `happy_breakdown` is `False` and `iteration_count < max_krylov_dim`, because I think your algo is wrong (cf my other comment on `break` after normal convergence).

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+        T[j + 1, j] = beta
+        T[j, j + 1] = beta
+
+        effective_dim = len(lanczos_vectors)
+        # increase size of T_truncated only when there is no happy breakdown
+        size = effective_dim + (0 if beta < norm_tolerance else 1)
+        T_truncated = T[:size, :size]
+
+        # create an initial guessed state for the eigensolver
+        if prev_eigen_vec is None:
```
pablolh, line 96: A suggestion instead of using None:
initialize `guessed_state` like here but before the loop
instead of storing `prev_eigen_vec` at each loop, update `guessed_state` when you can. So `guessed_state` becomes the variable that stays, not `prev_eigen_vec`
it's less messy IMO

emu_base/math/krylov_energy_min.py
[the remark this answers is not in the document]
elmerhej: I fixed that in the updated PR

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+            guessed_state = torch.zeros(
+                T_truncated.size(0), 1, dtype=dtype, device=device
+            )
+            guessed_state[0, 0] = 1.0
+        else:
+
+            if beta > norm_tolerance:
```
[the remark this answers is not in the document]
elmerhej: Done

emu_base/math/krylov_energy_min.py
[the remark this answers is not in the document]
elmerhej: Fixed that in updated PR

test/emu_base/math/test_krylov_energy_min.py
elmerhej: Done that. please see my Ising_ham test

emu_base/math/krylov_energy_min.py
elmerhej: I didn't look into that yet as I was mostly focused on implementing the second convergence criterion + writing new tests. I'll think about it next

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+        final_state = final_state / final_state.norm()
+
+        # residual norm ||A v - lambda * v|| convergence check
+        residual_norm = torch.norm(op(final_state) - ground_energy * final_state)
+        if residual_norm < residual_tolerance:
+            happy_breakdown = False
+            converged = True
+            iteration_count = j + 1
+            break
+        iteration_count = j + 1
```
pablolh, line 130: move this before `if`, it's in both branches

test/emu_base/math/test_krylov_energy_min.py
```diff
@@ -0,0 +1,192 @@
...
+dtype = torch.complex128
+
+
+def build_Ising_hamiltonian(N: int, J: float, h: float) -> torch.Tensor:
+    """Build the Ising Hamiltonian and test the Lanczos routine afterwards
+
+    H = -J * sum Z_i Z_{i+1} - h * sum X_i
+    Returns: H in matrix form
+    """
+    Identity = torch.eye(2, dtype=torch.float64)
```
pablolh, line 17: `identity`

test/emu_base/math/test_krylov_energy_min.py
```diff
@@ -0,0 +1,192 @@
...
+    X = torch.tensor([[0, 1], [1, 0]], dtype=torch.float64)
+    Z = torch.tensor([[1, 0], [0, -1]], dtype=torch.float64)
+
+    def construct_tensor_product(list_of_ops):
```
pablolh, line 21: Mauro wrote a function for this using `functools.reduce`, use it instead

test/emu_base/math/test_krylov_energy_min.py
```diff
@@ -0,0 +1,192 @@
...
+        res = list_of_ops[0]
+        for operator in list_of_ops[1:]:
+            res = torch.kron(res, operator)
+        return res
+
+    def build_operator(pauli_op, site):
```
pablolh, line 27: factorize this somehow with `def single_gate(i: int, nqubits: int, g: torch.Tensor):` from `test_hamiltonian.py`, for instance
Should we make a common testutils in emu_base?

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+
+        lanczos_vectors.append(w / beta)
+        prev_eigen_vec = ground_eigenvector.clone()
+        guessed_state = extend_initial_state(prev_eigen_vec, dtype, device)
+
+        # Reconstruct final state in original Hilbert space
+        final_state = sum(c * vec for c, vec in zip(ground_eigenvector, lanczos_vectors))
+        final_state = final_state / final_state.norm()
+
+        # residual norm ||A v - lambda * v|| convergence check
```
pablolh, line 123: careful with notations, `v` is the initial candidate not final state correct?

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+        ground_energy=ground_energy.item(),
+        converged=converged,
+        happy_breakdown=happy_breakdown,
+        iteration_count=iteration_count,
+    )
+
+
+def krylov_energy_minimization(
+    op: Callable[[torch.Tensor], torch.Tensor],
+    v: torch.Tensor,
```
pablolh, line 143: I feel bad about this `v` parameter name: the function's output shouldn't depend on it no? (contrary to the original krylov_exp)
I think this `v` could just be random and it is a "initial guess" or some kind of seed for the Krylov subspace.
What is it exactly? Is `v` a good name?

emu_base/math/krylov_energy_min.py
elmerhej: I completely agree! The notation in the comment should be something like:
||H * final_state - ground_energy * final_state|| according to my notations in the code
but I will remove it instead, since it is readable without it I think

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+    residual_tolerance: float,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Return the smallest (k = 1) eigenpair of T_trunc.
+    If T_trunc is too small for LOBPCG (n < 3), do a full eigh;
+    otherwise call lobpcg with the given initial guessed state.
+    """
+    n = T_trunc.size(0)
+    if n < 3:
+        return cast(Tuple[torch.Tensor, torch.Tensor], torch.linalg.eigh(T_trunc))
```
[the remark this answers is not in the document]
elmerhej: Yes, I just checked it. It works.
Done

emu_base/math/krylov_energy_min.py
elmerhej: Hmm, but what if the if statement is true? then iteration count = j + 1 will be added twice (once before as you are suggesting, and once inside the loop of the residual convergence check). Would it be correct?

or probably you're suggesting I remove iteration_count = j + 1 from both if statements? then I just add iteration_count = j + 1 before both if statements?

emu_base/math/krylov_energy_min.py
```diff
@@ -0,0 +1,164 @@
...
+        lanczos_vectors.append(w / beta)
+        prev_eigen_vec = ground_eigenvector.clone()
+        guessed_state = extend_initial_state(prev_eigen_vec, dtype, device)
+
+        # Reconstruct final state in original Hilbert space
+        final_state = sum(c * vec for c, vec in zip(ground_eigenvector, lanczos_vectors))
+        final_state = final_state / final_state.norm()
+
+        # residual norm ||A v - lambda * v|| convergence check
+        residual_norm = torch.norm(op(final_state) - ground_energy * final_state)
```
sgrava, line 124: `torch.dist(a,b)` is also an option

a-quelle: Looking at this here, but also a previous comment where Pablo suggested changing an inner product into a norm of difference makes me think: the norm of the difference is not a good distance measure on a projective space, because it treats the same vector with a different phase as different. Isn't it better to use (assuming both inputs are normalized)
`1-torch.abs(torch.inner(psi, phi))`

emu_base/math/krylov_energy_min.py
elmerhej: Yes, `v` is nothing but an initial guess for the Krylov subspace. Later on (after a more complete implementation of the dmrg algo) it will be the ground state solution of the previous 2-site MPS sites (at sites `i-1` and `i`) found in the previous dmrg iteration. 

I changed it to `v0` in this case, and removed the comment above. but we can also change it in the future when the dmrg implementation is complete, unless you have other suggestions?

test/emu_base/math/test_krylov_energy_min.py
elmerhej: Done, I used `single_gate` and the `sigma_x` matrix from `test_hamiltonian.py`
I agree that it would be nice to write a single common testutils in emu_base, let's see what the others think as well

pablolh (approved):
Elie, I don't think **Em/Lanczos_algorithm_ground_state_solver** is a good name for a commit

murogrande (approved):
Pablo approved this, I will trust him

PR #104, opened 2025-05-05
Fix releases off of different sources

pablolh:
Make sure that emu-sv and emu-mps are released from the right sources tagged with the release tag, and not the current state of main.

### DIFF
--- a/.github/workflows/publish_emu_mps.yml
+++ b/.github/workflows/publish_emu_mps.yml
@@ -7,7 +7,7 @@ on:
     types:
       - completed
 concurrency:
-  group: fast-${{ github.head_ref || github.run_id }}
+  group: fast-${{ github.event.workflow_run.head_sha || github.run_id }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_emu_mps.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Check emulators
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.event.workflow_run.head_sha }}
```
sgrava, line 25: Can we have at least a way to check?
I'm thinking about echoing or checking if the branch that triggered the release is the same.
Is it possible to compare these two SHA?

pablolh: That will be displayed in the logs, for instance, here is the checkout of the failed emu-mps publishing job
image

However, it seems complicated to run this job "for trial" before merging this PR.

The sha I changed to is the head of the triggering workflow.

sgrava (approved):

PR #109, opened 2025-05-07
Further reduce max memory usage thanks to TensorWrapper

pablolh:
See docstrings


emu_base/utils.py
```diff
@@ -7,3 +7,20 @@ def dist2(left: torch.Tensor, right: torch.Tensor) -> float:
...
+    that ignores the current reference count of the Tensor object.
+
+    It is useful when you want to free memory that is no longer used
+    inside a function but that memory is also owned by a variable
+    in the outer scope, making it impossible to free it otherwise.
+
+    After calling that function, the Tensor object
+    should no longer be used.
+    """
+    t.resize_(0)
```
pablolh, line 25: `resize_` is needed to avoid the warning

PR #113, opened 2025-05-09
add derivative respect to interaction matrix

### LINKED ISSUE #111: [emu-sv] gradient backpropagation for interaction matrix
In `EvolveStateVector.backward` add support for backpropagation of the gradient to the interaction matrix.

sgrava:
Related to #111 

First half of the changes to support backpropagation of gradients to atomic positions/register.
These changes cover the backpropagation from the gradient given by torch to the interaction matrix.

### Changes
- `EvolveStateVector.backward` also return the gradient with respect to the interaction matrix **if needed**.
- tested

The backpropagation from the interaction matrix to atomic position should be automatic, but currently the interaction matrix is not created in a differentiable way. I'll fix this in another PR.


emu_sv/time_evolution.py
```diff
@@ -276,8 +299,24 @@ def backward(ctx: Any, grad_state_out: torch.Tensor, gham: None) -> tuple[
                 v = dhp @ e_l
                 grad_phis[i] = (-1j * dt * torch.tensordot(Vg.conj(), v)).real
 
+        if ctx.needs_input_grad[4]:
+            grad_int_mat = torch.zeros_like(interaction_matrix)
+            for i in range(nqubits):
+                for j in range(i + 1, nqubits):
+                    dhu = DHDUSparse(i, j, e_l.device, nqubits)
+                    v = dhu @ e_l
+                    grad_int_mat[i, j] = (-1j * dt * torch.tensordot(Vg.conj(), v)).real
```
pablolh, line 308: is it not symmetric?

emu_sv/time_evolution.py
```diff
@@ -98,7 +98,27 @@ def __init__(self, index: int, device: torch.device, nqubits: int):
...
+        self.shape = (2**i, 2, 2 ** (j - i - 1), 2, 2 ** (nqubits - j - 1))
+        diag = torch.zeros(
+            *self.shape,
+            dtype=torch.complex128,
+            device=device,
+        )
+        diag[:, 1, :, 1] = 1.0
         self.diag = diag.reshape(-1)
 
     def __matmul__(self, vec: torch.Tensor) -> torch.Tensor:
```
pablolh, line 124: What this huge `self.diag` is implementing is just a binary mask, since it's only filled with 0s and 1s.
I think you don't need the extra memory for this, a better way is to just implement directly:
```
result = vec.clone()
result.reshape(self.shape)
result[:, 0, :, 0] = 0  # I think
return result
```
Let me know what you think.

pablolh: BTW I think you can apply similar technique to `DHDDeltaSparse`, zero out and then multiply by -1

emu_sv/time_evolution.py
sgrava: the other half is not used so it doesn't matter

emu_sv/time_evolution.py
sgrava: ah you are right, seems you can save one vector.
Please, have a look!

emu_sv/time_evolution.py
```diff
@@ -87,22 +87,19 @@ def __matmul__(self, vec: torch.Tensor) -> torch.Tensor:
...
+    def __init__(self, i: int, nqubits: int):
+        self.nqubits = nqubits
+        self.shape = (2**i, 2, 2 ** (nqubits - i - 1))
 
     def __matmul__(self, vec: torch.Tensor) -> torch.Tensor:
-        return vec * self.diag
+        result = vec.clone()
+        result = result.reshape(vec.shape[0], *self.shape)
+        result[:, :, 0] = 0.0
+        result[:, :, 1] *= -1.0
```
pablolh, line 101: can't you just multiply the whole of `result`? since 0 * x == 0
`result *= -1.0`
not completely sure what is better but I think avoiding the slicing might actually be more efficient

emu_sv/time_evolution.py
pablolh: looks good

test/utils_testing/utils_interaction_matrix.py
```diff
@@ -1,15 +1,25 @@
 import torch
 
 
-def randn_interaction_matrix(N: int):
-    temp_mat = torch.randn(N, N).fill_diagonal_(0.0)
-    interaction = (temp_mat + temp_mat.mT) / 2
```
kbidzhiev, line 6: Stef, why did you prefer for-loop instead of `randn(N, N)` ?

sgrava: @kbidzhiev nothing against that!
Unfortunately, `fill_diagonal_(0.0)` is not compatible with differentiability.

PR #116, opened 2025-05-14
fix monkeypatch on save

a-quelle:
The `get_state` method of `MPSBackendImpl` destoys the monkeypatching of the observables. This expresses itself during the checkpointing.

### DIFF
--- a/emu_mps/mps_backend_impl.py
+++ b/emu_mps/mps_backend_impl.py
@@ -170,8 +170,10 @@ class MPSBackendImpl:
 
     def __getstate__(self) -> dict:
         d = self.__dict__.copy()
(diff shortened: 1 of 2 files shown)


kbidzhiev:
Should you then remove conversion to `torch.tensor if list` that you did in a part where optimatrix permutes the result back

a-quelle:
> Shouldn't you then remove conversion to `torch.tensor if list` that you did in a part where optimatrix permutes the result back

You might still need this if you somehow decide to serialize and unserialize a results object, since converting it to json converts all `torch.Tensor` objects to `list`. I agree that it's unlikely to trigger, but it won't rot either.

kbidzhiev (approved):

PR #118, opened 2025-05-14
[emulators] Sg/115 emulators make interaction matrix differentiable again

### LINKED ISSUE #115: [emulators] make interaction matrix differentiable again
Currently, the interaction matrix is not differentiable. 
At a first inspection, the culprit seems to be the utils.dist2/3 functions, which return a float instead of a `torch.tensor`.

### Acceptance criteria
- interaction matrix is differentiable
- small test to fix the behavior

sgrava:
resolve #115

### Changes
- force qubit positions to have type `torch.float64`
- force `interaction_matrix` to have type `torch.float64`
- new test to cover xy interaction when it is actively used to cancel some interaction
 
In making so, one end-to-end test in emu_sv started to fail. 
In my defense, it is a debatable test that just fix the end value energy after some sequence, up to some tolerance.

### Minor changes
- took the occasion to reshape the interaction matrix creation code and the docstrings

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,43 @@
+cff-version: 1.2.0
+title: pasqal emulators
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: Kemal
+    family-names: Bidzhiev
+    email: kemal.bidzhiev@pasqal.com
+    affiliation: Pasqal
+  - email: stefano.grava@pasqal.com
+    given-names: Stefano
+    family-names: Grava
+    affiliation: Pasqal
+  - given-names: Pablo
+    name-particle: le
+    family-names: Henaff
+    email: pablo.le-henaff@pasqal.com
+    affiliation: Pasqal
+  - given-names: Mauro
+    family-names: Mendizabel Pico
+    email: mauro.mendizabal-pico@pasqal.com
+    affiliation: Pasqal
+  - given-names: Elie
+    family-names: Merhej
+    email: elie.merhej@pasqal.com
+    affiliation: Pasqal
+  - given-names: Anton
+    family-names: Quelle
+    email: anton.quelle@pasqal.com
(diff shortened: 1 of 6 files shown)


murogrande:
regarding the tests in emu-sv, previously the interaction matrix had dtype= torch.float32, now it is torch.float64 and the error is in the 5th decimal, I can say this is expected for float32

- torch.float32: single precision has around 7 decimal digits of precision.
- torch.float64: double precision has around 15 decimal digits of precision.

a-quelle (approved):
lgtm

a-quelle (approved):
lgtm

emu_base/pulser_adapter.py
```diff
@@ -23,67 +20,59 @@ def _get_qubit_positions(
...
-                / dist3(qubit_positions[numi], qubit_positions[numj])
-            )
-            interaction_matrix[numj, numi] = interaction_matrix[numi, numj]
-
+    mag_field = torch.tensor(sequence.magnetic_field, dtype=torch.float64)
+    mag_field /= mag_field.norm()
+    positions = _get_qubit_positions(sequence.register)
+
+    interaction_matrix = torch.zeros(nqubits, nqubits, dtype=torch.float64)
+    for i in range(nqubits):
```
pablolh, line 71: what about using a version of this code from @kbidzhiev  instead:
```
def reciprocal_dist_matrix(reg: pulser.Register) -> torch.Tensor:
    """
    Matrix 1/r_{ij}. Behaves similarly as 1/r_{ij}^6
    """
    positions = torch.tensor([q.tolist() for q in reg.qubits.values()])

    diff = positions.unsqueeze(1) - positions.unsqueeze(0)
    dist = (diff**2).sum(-1).pow(0.5)

    mask = dist != 0
    dist_inv = torch.zeros_like(dist)
    dist_inv[mask] = 1.0 / dist[mask]

    return dist_inv
```

PR #119, opened 2025-05-15
[emu-sv] connecting pulser with master eq solver 

murogrande:
Pulser `NoiseModel` with effective noise channels  is accepted by emu-sv. 
 - The connection with pulser was made using a similar `backend_impl` in emu-mps. This means that the krylov solver changes some parameters according if there a noise simulation or not. 
 - These parameters are: the `__matmat__` @ in the case of density matrix with `RydebergLindbladian` class, `is_hermitian` flag will be off. 

Also, I added some tests. The most important is the `end_to_end_spontaneous_emission_rate` one, where I tested against pulser master equation solver.


emu_sv/custom_callback_implementations.py
```diff
@@ -34,6 +34,38 @@ def qubit_occupation_sv_impl(
...
+def qubit_occupation_sv_den_mat_impl(
+    self: Occupation,
+    *,
+    config: EmulationConfig,
+    state: DensityMatrix,
+    hamiltonian: DenseOperator,
+) -> torch.Tensor:
+    """
+    Custom implementation of the occupation nᵢ observable for density matrix.
+    The observable nᵢ is given by: I ⊗ ... ⊗  nᵢ ⊗ I ⊗I ⊗
```
kbidzhiev, line 46: the last kroneker ⊗ is redundant

emu_sv/custom_callback_implementations.py
```diff
@@ -34,6 +34,38 @@ def qubit_occupation_sv_impl(
...
+    where nᵢ is the occupation operator for qubit i.
+    The expectation value is given by: <nᵢ> = Tr(ρ nᵢ).
+
+    The output will be a tensor of size (nqubits,), where each element will be the
+    expectation value of the occupation operator for each qubit.
+    In case of 3 atoms, the output will be a tensor of size (3,), where each element
+    will be <nᵢ> = Tr(ρᵢnᵢ), or [ <n₁>, <n₂>, <n₃> ].
```
kbidzhiev, line 53: ρᵢ is just ρ with no index

emu_sv/custom_callback_implementations.py
```diff
@@ -34,6 +34,38 @@ def qubit_occupation_sv_impl(
...
+    """
+    nqubits = state.n_qudits
+    occupation = torch.zeros(nqubits, dtype=torch.float64, device=state.matrix.device)
+    for i in range(nqubits):
+        state_tensor = (
+            state.matrix.view(
+                2**i, 2, 2 ** (nqubits - i - 1), 2**i, 2, 2 ** (nqubits - i - 1)
+            )[:, 1, :, :, 1, :]
+            .contiguous()
```
kbidzhiev, line 62: are you sure you need `.contiguous()` ? it returns a tensor

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,243 @@
...
+                self.target_times[step + 1] / self.target_times[-1],
+                state,
+                H,  # type: ignore[arg-type]
+                self.results,
+            )
+            self.time = time.time()
+            del H
+
+
+class NoyseSVBackendImpl(SVBackendImpl):
```
kbidzhiev, line 151: do you mean "noise" ?

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,243 @@
...
+    Args:
+        sequence: The sequence to be emulated.
+        config: The configuration for the emulator.
+
+    Returns:
+        An instance of SVBackendImpl.
+    """
+    pulse_data = PulserData(sequence=sequence, config=config, dt=config.dt)
+    if pulse_data.has_lindblad_noise:
+        return NoyseSVBackendImpl(config, pulse_data)
```
kbidzhiev, line 241: Noise noyse

emu_sv/sv_config.py
```diff
@@ -115,21 +117,40 @@ def _expected_kwargs(self) -> set[str]:
...
+                    obs_copy.apply = MethodType(  # type: ignore[method-assign]
+                        qubit_occupation_sv_den_mat_impl, obs_copy
+                    )
+            if isinstance(obs, EnergyVariance):
+                if self.noise_model.noise_types == ():
+                    obs_copy.apply = MethodType(  # type: ignore[method-assign]
+                        energy_variance_sv_impl, obs_copy
+                    )
+                else:
+                    raise Exception("Not implemented for density matrix")
```
kbidzhiev, line 139: you can make these blocks shorter with smth like:
```python
if self.noise_model.noise_types != ():
  raise Exception("Not implemented for density matrix")
obs_copy.apply = MethodType(  # type: ignore[method-assign]
                        energy_variance_sv_impl, obs_copy
                    )
```

because `raise` interrupts the further code execution

emu_sv/custom_callback_implementations.py
murogrande: yes, let me change this

emu_sv/custom_callback_implementations.py
murogrande: yes, let me change it

emu_sv/sv_config.py
```diff
@@ -115,21 +117,37 @@ def _expected_kwargs(self) -> set[str]:
...
+
             if isinstance(obs, Occupation):
-                obs_copy.apply = MethodType(  # type: ignore[method-assign]
-                    qubit_occupation_sv_impl, obs_copy
-                )
-            elif isinstance(obs, EnergyVariance):
+                if self.noise_model.noise_types == ():
+                    obs_copy.apply = MethodType(  # type: ignore[method-assign]
+                        qubit_occupation_sv_impl, obs_copy
+                    )
```
kbidzhiev, line 128: what do you think about:

```python
if self.noise_model.noise_types:
    raise NotImplementedError( ...)
```

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,243 @@
...
+                self.target_times[step + 1] / self.target_times[-1],
+                state,
+                H,  # type: ignore[arg-type]
+                self.results,
+            )
+            self.time = time.time()
+            del H
+
+
+class NoisySVBackendImpl(SVBackendImpl):
```
pablolh, line 151: There is no point in doing inheritance here, it seems.
There are no shared methods, and no shared data members. 

If you want to use inheritance, please call `super().__init()` in the constructor of the base class to factorize code.

Then, you need to create a new method like `run()` in the base class to distinguish. Something like `fill_results` also may help.

Copy-pasting existing code is BAD. Every time you do it, it's the sign there is something wrong in terms of software design.

All this piece should not occur more than once, for instance:
```
        self._config = config
        self._pulser_data = pulser_data
        self.target_times = pulser_data.target_times
        self.time = 0.0
        self.statistics = None
        self.omega = pulser_data.omega
        self.delta = pulser_data.delta
        self.phi = pulser_data.phi
        self.nsteps = pulser_data.omega.shape[0]
        self.nqubits = pulser_data.omega.shape[1]
```

pablolh: (there are more shared pieces, I will let you find them and factorize)

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,243 @@
...
+    results: Results
+
+    def __init__(self, config: SVConfig, pulser_data: PulserData):
+        """
+        Initializes the SVBackendImpl.
```
pablolh, line 156: Docstring is wrong

pablolh (commented):
How does it play with differentiability (now possible in emu-sv without noise @sgrava )? Does it work? If not it would be good to have a check and an error with a clear message.

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+_TIME_CONVERSION_COEFF = 0.001  # Omega and delta are given in rad/μs, dt in ns
+
+
+class SVBackendImpl:
+
+    results: Results
+
+    def __init__(self, config: SVConfig, pulser_data: PulserData):
```
a-quelle, line 78: I think we can share the code better between `SVBackendImpl` and `NoisySVBackendImpl`. For example, the lines up to here are identical in both `init` methods, so is the initialization of `self.results` and `self.statistics` and everything in the for loop, except for the function that does the time-step.

The parts that are not shared, are how the initial state is created, and what stepper is used. You could get those from methods `set_initial_state` and `get_stepper` which could have different implementations in both classes. Then you can share all the other code by using these functions.

Additionally, I think it's nicer if the for loop is in a `run` method.

a-quelle (changes requested):

emu_sv/sv_backend_impl.py
sgrava: @MauroMendizabal @pablolh no need to worry about differentiability here.
Despite the not noisy sv implementation being differentiable, there is no API for a user to make use of it. It will be more convenient to add the logic `if noise do something else do something else` there I think.

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,264 @@
...
+        self.pulser_lindblads = pulser_data.lindblad_ops
+
+    def initial_state(self) -> State:
+        if self._config.initial_state is not None:  # fix this with state vector
+            state = self._config.initial_state
+            return DensityMatrix(state.matrix.clone(), gpu=self._config.gpu)
+        else:
+            return DensityMatrix.make(self.nqubits, gpu=self._config.gpu)
+
+    def _run(self) -> Results:
```
pablolh, line 199: You still have two `run` methods which look extremely similar, and are quite long.
I think it is possible to have a single `run` method.

You could create a `BaseImpl` class if you want to make this nicer. Think of `BaseImpl` as an interface implementing what is common to all implementations. In that case, both implementations iterate steps with a dt, fill in results and statistics.

It would be for me a criterion for a good OOP design to have a SINGLE loop calling callbacks.

So,
`class Base`, `class SVBackendImpl(Base)` and `class DensityMatrixBackendImpl(Base)`

What you need to understand is that a _base class implementation can call abstract methods implemented in the derived classes_, and the behavior will depend on the dynamic type of the class. Think of this behavior as "do something based on what you are, I don't know yet".

Of course it's easier said than done. For instance it seems you have this `self.pulser_lindblads` argument that is different between stepper of SV and stepper of DM. How to deal with this? There are multiple ways.

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+
+    def initial_state(self) -> State:
+        if self._config.initial_state is not None:
+            state = self._config.initial_state
+            return StateVector(state.vector.clone(), gpu=state.vector.is_cuda)
+        else:
+            return StateVector.make(self.nqubits, gpu=self._config.gpu)
+
+    def _choose_stepper(self, state: State) -> typing.Callable:
```
pablolh, line 111: The body of this function is the symptom that there is something wrong with your design: `SVBackendImpl` shouldn't mention `EvolveDensityMatrix` at all.

See my comment below where I suggest creating a `Base`

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,263 @@
...
+    Returns:
+        An instance of SVBackendImpl.
+    """
+    pulse_data = PulserData(sequence=sequence, config=config, dt=config.dt)
+    backend: SVBackendImpl
+    if pulse_data.has_lindblad_noise:
+        backend = NoisySVBackendImpl(config, pulse_data)
+    else:
+        backend = SVBackendImpl(config, pulse_data)
+    backend._run()
```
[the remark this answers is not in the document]
a-quelle: agreed

emu_sv/sv_backend_impl.py
a-quelle: Yes, `_choose_stepper` should always return `EvolveStateVector` in the non-noisy class. In the noisy class, you can reimplement this method and always return `EvolveDensityMatrix.evolve`, and then they become nicely separated.

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,264 @@
...
+    def initial_state(self) -> State:
+        if self._config.initial_state is not None:  # fix this with state vector
+            state = self._config.initial_state
+            return DensityMatrix(state.matrix.clone(), gpu=self._config.gpu)
+        else:
+            return DensityMatrix.make(self.nqubits, gpu=self._config.gpu)
+
+    def _run(self) -> Results:
+
+        state = self.initial_state()
```
a-quelle, line 201: there's a `self.time=time()` missing here. Other than that, the only difference I can see between the two run methods is choosing the stepper and initializing the state. Both these things can be done in the `init` that way you only need a single `run` method.

emu_sv/sv_config.py
a-quelle: I don't think this comment is at quite the right line of code, but Kemal has a valid point. Somewhere in the `SVConfig` there should be a check that only effective noise is present in the `NoiseModel`

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,244 @@
...
+            self.results,
+        )
+        self.time = time.time()
+        del self._current_H
+
+    def _run(self) -> None:
+        for step in range(self.nsteps):
+            self.step(step)
+
+        # NOTE: return self.results or not?
```
pablolh, line 149: could be a good idea indeed

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,244 @@
...
+        self._pulser_data = pulser_data
+        self.target_times = pulser_data.target_times
+        self.omega = pulser_data.omega
+        self.delta = pulser_data.delta
+        self.phi = pulser_data.phi
+        self.nsteps = pulser_data.omega.shape[0]
+        self.nqubits = pulser_data.omega.shape[1]
+        self.time = time.time()
+        self.state = self.initial_state()
+        self.stepper: typing.Callable
```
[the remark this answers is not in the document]
murogrande: the stepper is important for the different `evolve` functions.  Now we the `class EvolveDensityMatrix` which has an `evolve` function for the noise case and there is a `EvolveStateVector` which has an `evolve` function. 
I can not remove it from the `DensityMatrixImpl`. 

I delete this fromt he base class but not from the `DensityMatrixImpl`.

emu_sv/sv_backend.py
```diff
@@ -1,23 +1,14 @@
...
-
-
-_TIME_CONVERSION_COEFF = 0.001  # Omega and delta are given in rad/μs, dt in ns
+from emu_sv.sv_backend_impl import create_impl
 
 
 class SVBackend(EmulatorBackend):
     """
     A backend for emulating Pulser sequences using state vectors and sparse matrices.
+    Noysi simulation is supported by solving the Lindblad equation and using effective
```
pablolh, line 10: Noisy

emu_sv/sv_backend_impl.py
```diff
@@ -0,0 +1,247 @@
...
+            if self._config.initial_state is None
+            else DensityMatrix(
+                self._config.initial_state.matrix.clone(), gpu=self._config.gpu
+            )
+        )
+
+        self.stepper = EvolveDensityMatrix.evolve
+
+    def _evolve_step(self, dt: float, step_idx: int) -> None:
+        self.state.matrix, self._current_H = self.stepper(
```
pablolh, line 220: what I meant with removing self.stepper  is directly calling `EvolveDensityMatrix.evolve` here

pablolh (dismissed):
for me is good

a-quelle (approved):
lgtm

murogrande:
Main points: 
- noisy simulations can be run from pulser. LIndblad master equation solver is integrated with pulser
- Support observables: occupation, bitstrings and state. The rest is not working yet (Future work)
- BaseSVBanckend manages noisy and non noisy simulations.
- ent_to_end tests can run with gpu, bitstrings and energy variance will fail
- bug: initial state number of atoms is different from the sequence number of atoms.
A new test that checks this was added

PR #124, opened 2025-05-23
Minimize energy of the pair state

elmerhej:
- method minimize_energy_pair() that minimizes the energy of a consecutive MPS sites using the Lanczos/Arnoldi algorithm. 
-  a unittest that accounts for above
- fix in the Lanzcos algorithm to account for non-hermitian matrices

### DIFF
--- a/emu_base/math/krylov_energy_min.py
+++ b/emu_base/math/krylov_energy_min.py
@@ -8,23 +8,33 @@ def _eigen_pair(
     T_trunc: torch.Tensor,
     guessed_state: torch.Tensor,
     residual_tolerance: float,
+    is_hermitian: bool,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     """
-    Return the smallest (k = 1) eigenpair of T_trunc.
-    If T_trunc is too small for LOBPCG (n < 3), do a full eigh;
-    otherwise call lobpcg with the given initial guessed state.
+    Return the smallest eigenpair of T_trunc.
+    If T_trunc is small (n < 3) or Hermitian, do a full eigh;
+    otherwise use lobpcg with the given initial guessed state.
     """
     n = T_trunc.size(0)
-    if n < 3:
-        return tuple(torch.linalg.eigh(T_trunc))
+    if is_hermitian:
+        if n > 2:
+            return torch.lobpcg(
+                T_trunc,
+                k=1,
+                X=guessed_state,
+                tol=residual_tolerance,
+                largest=False,
+            )
+        else:
+            return tuple(torch.linalg.eigh(T_trunc))
 
-    return torch.lobpcg(
-        T_trunc,
-        k=1,
-        X=guessed_state,
-        tol=residual_tolerance,
-        largest=False,
-    )
+    # eig does not guarantee that eigenvalues are sorted
+    # use argmin to force extract the smallest eigen-pair
+    eigvals, eigvecs = torch.linalg.eig(T_trunc)
+    idx = torch.argmin(eigvals.abs())
+    energy = eigvals[idx : idx + 1]
+    state = eigvecs[:, idx : idx + 1]
+    return energy, state
 
 
 def extend_initial_state(
@@ -104,8 +114,10 @@ def krylov_energy_minimization_impl(
         if prev_eigen_vec is None:
             guessed_state = torch.ones(size, 1, dtype=dtype, device=device)
 
-        eigvals, eigvecs = _eigen_pair(T_truncated, guessed_state, residual_tolerance)
-        ground_energy = eigvals[0].real
+        eigvals, eigvecs = _eigen_pair(
+            T_truncated, guessed_state, residual_tolerance, is_hermitian
+        )
+        ground_energy = eigvals[0]
         ground_eigenvector = eigvecs[:, 0]  # in Krylov subspace
         iteration_count = j + 1
 
--- a/emu_mps/dmrg.py
+++ b/emu_mps/dmrg.py
@@ -1,0 +1,84 @@
+import torch
+
+from emu_base.math.krylov_energy_min import krylov_energy_minimization
+from emu_base.utils import deallocate_tensor
+from emu_mps import MPSConfig
+from emu_mps.utils import split_tensor
+from emu_mps.tdvp import apply_effective_Hamiltonian
+
+
+def minimize_energy_pair(
+    *,
+    state_factors: list[torch.Tensor],
+    baths: tuple[torch.Tensor, torch.Tensor],
+    ham_factors: list[torch.Tensor],
+    orth_center_right: bool,
+    is_hermitian: bool,
+    config: MPSConfig,
+    residual_tolerance: float,
+) -> tuple[torch.Tensor, torch.Tensor, float]:
+    """
+    Minimizes the state factors (ψ_i, ψ_{i+1}) using the Lanczos/Arnoldi method
+    """
+
+    assert len(state_factors) == 2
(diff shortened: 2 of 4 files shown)


emu_base/math/krylov_energy_min.py
```diff
@@ -8,23 +8,33 @@ def _eigen_pair(
...
+                largest=False,
+            )
+        else:
+            return tuple(torch.linalg.eigh(T_trunc))
+
+    # eig does not guarantee that eigenvalues are sorted
+    # use argmin to force extract the smallest eigen-pair
+    eigvals, eigvecs = torch.linalg.eig(T_trunc)
+    idx = torch.argmin(eigvals.abs())
+    energy = eigvals[idx : idx + 1]
```
kbidzhiev, line 35: eigvals is a list. Wouldn't just `eigvals[idx]` work ?

emu_base/math/krylov_energy_min.py
```diff
@@ -8,23 +8,33 @@ def _eigen_pair(
...
+    state = eigvecs[:, idx : idx + 1]
```
kbidzhiev, line 36: the same
```
state = eigvecs[:, idx]
```
?

emu_base/math/krylov_energy_min.py
```diff
@@ -8,23 +8,33 @@ def _eigen_pair(
...
     residual_tolerance: float,
+    is_hermitian: bool,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     """
-    Return the smallest (k = 1) eigenpair of T_trunc.
-    If T_trunc is too small for LOBPCG (n < 3), do a full eigh;
-    otherwise call lobpcg with the given initial guessed state.
+    Return the smallest eigenpair of T_trunc.
+    If T_trunc is small (n < 3) or Hermitian, do a full eigh;
+    otherwise use lobpcg with the given initial guessed state.
```
kbidzhiev, line 16: change a bit the description. The code uses `lobpcg` for large hermitean matrices.

It seems that it will be good to join what `eigh` and `eig` do here. `eigh` vs `eig` performance gain for `n==2` is not significant

emu_base/math/krylov_energy_min.py
```diff
@@ -8,23 +8,33 @@ def _eigen_pair(
...
+    if is_hermitian:
+        if n > 2:
+            return torch.lobpcg(
+                T_trunc,
+                k=1,
+                X=guessed_state,
+                tol=residual_tolerance,
+                largest=False,
+            )
+        else:
```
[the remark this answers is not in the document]
a-quelle: kemal is right, just doing
`if is_hermitian and n > 1: return torch.lobpgc` and skipping the else is good enough. For a 2x2 matrix the difference between `eigh` and `eig` will not be noticable. It will make the code simpler

emu_mps/dmrg.py
```diff
@@ -0,0 +1,85 @@
...
+from emu_base.math.krylov_energy_min import krylov_energy_minimization
+from emu_base.utils import deallocate_tensor
+from emu_mps import MPSConfig
+from emu_mps.utils import split_tensor
+from emu_mps.tdvp import apply_effective_Hamiltonian
+
+
+def minimize_energy_pair(
+    *,
+    state_factors: list[torch.Tensor],
```
a-quelle, line 12: make it a tuple?

elmerhej: Done

emu_base/math/krylov_energy_min.py
```diff
@@ -8,23 +8,33 @@ def _eigen_pair(
     T_trunc: torch.Tensor,
     guessed_state: torch.Tensor,
     residual_tolerance: float,
+    is_hermitian: bool,
```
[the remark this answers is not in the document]
elmerhej: I made major changes in the code after a discussion with @a-quelle  and @kbidzhiev , therefore the decision is to remove the:
1) non-hermitian feature as DMRG's core idea is to minimize a hermitian tensor (H) to extract the energy (real valued) of the system
2) swap the torch.lobpcg with the linalg eigh operation

Therefore we won't be needing these args anymore (guessed_state, residual_tol, is_hermitian etc.) in _eigen_pair()

emu_mps/dmrg.py
```diff
@@ -0,0 +1,85 @@
...
+    right_bath = right_bath.to(left_device)
+
+    combined_hamiltonian_factors = (
+        torch.tensordot(left_ham_factor, right_ham_factor, dims=1)
+        .transpose(2, 3)
+        .contiguous()
+        .view(left_ham_factor.shape[0], 4, 4, -1)
+    )
+
+    op = lambda x: apply_effective_Hamiltonian(
```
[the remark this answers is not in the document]
elmerhej: I don't think there is major code duplication here, because the presented code is calling 'apply_effective_hamiltonian' from ```tdvp.py```, and hence ```op``` object is constructed (note that ```op``` here differs from the ``op`` constructed in TDVP, by the phase factor -1j \times dt so it's not exactly the same object here and there).

However, if you think we can generalize it a bit further, I have the following suggestion:

```
def make_op(
    *,
    combined_hamiltonian_factors,
    left_bath,
    right_bath,
    evolution_type,
    dt=None,
):
    if evolution_type == EvolutionType.KRYLOV_EXP:
        if dt is None:
            raise ValueError("dt is required for KRYLOV_EXP")
        phase = -_TIME_CONVERSION_COEFF * 1j * dt
        return lambda x: phase * apply_effective_Hamiltonian(
            x, combined_hamiltonian_factors, left_bath, right_bath
        )
    else:  
        return lambda x: apply_effective_Hamiltonian(
            x, combined_hamiltonian_factors, left_bath, right_bath
        )
```
what do you think?

one-site DMRG is fundamentally different than two-site DMRG (you can think of it like the fundamental differences between one-site tdvp where \chi is constant from the start, and our current 2-site tdvp routine). In conclusion, we only implement a two-site dmrg here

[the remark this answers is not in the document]
a-quelle: how about
```
def make_op(time_step:float)
    assert len(state_factors) == 2
    assert len(baths) == 2
    assert len(ham_factors) == 2

    left_state_factor, right_state_factor = state_factors
    left_bath, right_bath = baths
    left_ham_factor, right_ham_factor = ham_factors

    left_device = left_state_factor.device
    right_device = right_state_factor.device

    left_bond_dim = left_state_factor.shape[0]
    right_bond_dim = right_state_factor.shape[-1]

    # Computation is done on left_device (arbitrary)

    right_state_factor = right_state_factor.to(left_device)

    combined_state_factors = torch.tensordot(
        left_state_factor, right_state_factor, dims=1
    ).reshape(left_bond_dim, 4, right_bond_dim)

    deallocate_tensor(left_state_factor)
    deallocate_tensor(right_state_factor)

    left_ham_factor = left_ham_factor.to(left_device)
    right_ham_factor = right_ham_factor.to(left_device)
    right_bath = right_bath.to(left_device)

    combined_hamiltonian_factors = (
        torch.tensordot(left_ham_factor, right_ham_factor, dims=1)
        .transpose(2, 3)
        .reshape(left_ham_factor.shape[0], 4, 4, -1)
    )

    op = (
        lambda x: time_step
        * apply_effective_Hamiltonian(
            x, combined_hamiltonian_factors, left_bath, right_bath
        )
    )
```

and then you call this with a `time_step` of `1.` for `minimize_energy_pair` and `-TIME_CONVERSION_COEFF * 1j * dt` for `evolve_pair`

PR #125, opened 2025-05-27
doppler noise

a-quelle:
Support for doppler noise.
I also refactored the pulser adapter and the tests a bit to make the entire affair more modular.


emu_base/pulser_adapter.py
```diff
@@ -137,49 +205,10 @@ def _extract_omega_delta_phi(
...
-                if i in times_to_amp_factors:  # multiple local channels at the same time
-                    # pulser enforces that no two lasers target the same qubit simultaneously
-                    # so only a single factor will be != 1.0 for each qubit
-                    times_to_amp_factors[i] = factors * times_to_amp_factors[i]
-                else:
-                    times_to_amp_factors[i] = factors
+    times_to_amp_factors = _get_amp_factors(
+        samples, amp_sigma, laser_waist, qubit_positions, q_ids
+    )
+    doppler_offset = _get_delta_offset(len(q_ids), temperature)
```
pablolh, line 211: nitpicking on this important code style guideline

> Ideally, declare and define each variable close to where it’s used.

so this line could be moved below

test/emu_base/test_adapter.py
```diff
@@ -339,10 +357,59 @@ def test_interaction_matrix_differentiability():
...
+        assert torch.allclose(val, expected_amp_factors[key] * waist_amplitudes)
+
+
+@patch("emu_base.pulser_adapter.random_gaussian")
+def test_get_delta_offset(mock_random):
+    mock_random.return_value = torch.tensor([1.0, 2.0, 3.0])
+    actual = _get_delta_offset(3, 5.0)
+    sigma = 0.18978408784721654
+    mock_random.assert_called_once_with((3,), 0.0, sigma)
+    assert torch.allclose(actual, mock_random.return_value, rtol=0.0, atol=0.0)
```
pablolh, line 390: `allclose` with 0 tolerance?

emu_base/pulser_adapter.py
```diff
@@ -77,13 +81,77 @@ def _xy_interaction(sequence: pulser.Sequence) -> torch.Tensor:
...
+    return times_to_amp_factors
+
+
+def _get_delta_offset(nqubits: int, temperature: float) -> torch.Tensor:
+    if temperature == 0.0:
+        return torch.zeros(nqubits, dtype=torch.float64)
+    else:
+        t = temperature * 1e-6  # microKelvin -> Kelvin
+        sigma = KEFF * math.sqrt(KB * t / RUBIDIUM_MASS)
+        return random_gaussian((nqubits,), 0.0, sigma)
```
sgrava, line 144: Just a suggestion. You could frame this in function of `det_sigma` .
This way:
- similar code to `amp_sigma`
- you compute `det_sigma` beforehand in the code, since it just depends on hardware specs.

On the docstring side I would add
 - https://pulser.readthedocs.io/en/stable/tutorials/laser_noise.html#Doppler-shift
 - a simple formula like $\sigma_{\Delta}=k_{eff}\sigma_{v}$, where $k_eff$ is the norm of the wavevector of the trapping laser

emu_base/pulser_adapter.py
a-quelle: done

emu_base/pulser_adapter.py
```diff
@@ -76,12 +81,77 @@ def _xy_interaction(sequence: pulser.Sequence) -> torch.Tensor:
...
+                    times_to_amp_factors[i] = factors * times_to_amp_factors[i]
+                else:
+                    times_to_amp_factors[i] = factors
+    return times_to_amp_factors
+
+
+def _get_delta_offset(nqubits: int, temperature: float) -> torch.Tensor:
+    if temperature == 0.0:
+        return torch.zeros(nqubits, dtype=torch.float64)
+    else:
```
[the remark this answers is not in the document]
a-quelle: indeed

emu_base/pulser_adapter.py
sgrava, line 144: Unrelated to this MR but I found it randomly (no pun intended) and looks like a nice solution: [torch.normal](https://docs.pytorch.org/docs/stable/generated/torch.normal.html)

```python
>>> omega = torch.ones(3)
>>> noisy_omega = torch.normal(omega, std=0.1)
>>> noisy_omega
tensor([1.0004, 0.9449, 1.0662])
```

a-quelle: regarding your final comment, the solution is to not support doppler noise for the XY channel. I tackled this issue in a systematic manner for all noise types in pulser adapter.

I wrote a docstring showing where the delta offset comes from.

Regarding your suggestion for `det_sigma`, I don't really like this, the parameter in the noise model is `temperature` and to convert it into `det_sigma` in a separate function just increases the size of the code, while not adding any serious benefit in terms of modularity and testability.

sgrava (approved):

emu_base/pulser_adapter.py
```diff
@@ -77,13 +100,83 @@ def _xy_interaction(sequence: pulser.Sequence) -> torch.Tensor:
...
+    The delta values are shifted due to atomic velocities.
+    The atomic velocities follow the Maxwell distribution
+    https://en.wikipedia.org/wiki/Maxwell%E2%80%93Boltzmann_distribution
+    and then a given residual velocity is converted to a delta offset per
+    https://en.wikipedia.org/wiki/Doppler_broadening
+    """
+    if temperature == 0.0:
+        return torch.zeros(nqubits, dtype=torch.float64)
+    t = temperature * 1e-6  # microKelvin -> Kelvin
+    sigma = KEFF * math.sqrt(KB * t / RUBIDIUM_MASS)
```
kbidzhiev, line 168: I'm a bit surprised with these small numbers. Dont you think that it's more safe to introduce a ration `KEFF*sqrt(KB/RUBIDIUM_MASS)` instead of separate entities

a-quelle: I merged `KB` and `RUBIDIUM_MASS` into `KB_PER_RUBIDIUM_MASS`, and it doesn't appreciably change anything. I've kept the change because I think it's alright, but I think merging it further would destroy understanding of where the constants come from, without adding numerical benefit.

kbidzhiev (approved):

PR #128, opened 2025-05-29
[emu-sv] restore differentiability

### LINKED ISSUE #127: [emu-sv] restore differentiability
Refactoring of the `SVBackend_impl` broke differentiability and managed to get around a test.
 
### Goal
Restore differentiability and specialize test.

### Description

Latest refactor
https://github.com/pasqal-io/emulators/blob/32180897e64f205b78d15513b501a93328acdfe7/emu_sv/sv_backend_impl.py#L174-L178

Introduced this logic to use or not the `autograd.Function` derived stepper in the time evolution.
This was done to separate the two workflows:
1. I will need to compute gradients in the future, so let torch do its magic.
2. I don't need gradient, so just compute the step in the normal way.

While separating the two workflows conceptually nice, in both cases torch will build a computational graph if some Hamiltonian parameter is flagged with require grad.
However, in 2, it will use its built methods. For this reason, when torch spots that we are applying `index_add_`, it will complain since its automatic differentiation does not support such operation. It is for this exact reason that we wrote our own backpropagation which goes around this issue.

Overall, probably the point is that `initial_state.requires_grad` is not sufficient to rule out that you'll need some backpropagation. The only way that comes to my mind is to assert that **all** parameters omega, delta, phi, interactions, initial state, do not require grad. But torch probably do this automatically.

### Actions
- [  ] better test
- [  ] remove logic `if initial_state.requires_grad`
- [  ] check again that torch overhead for running forward without gradient requirements is negligible

sgrava:
resolve #127 

### Changes
- partially in-place krylov_exp (can be improved but let's do it in another MR)
- remove faulty logic `if initial_state.requires_grad`, see issue description #127 
- remove old test. **krylov_exp is no longer differentiable** per se,  one must always define a forward/backward 

Following the issue description #127, we introduce two new tests, which highlight two issues
1. Hamiltonian application (index_add) is not called in the right `no_grad` context in forward
```python
E           TypeError: index_add_(): argument 'alpha' must be Number, not Tensor

emu_sv/hamiltonian.py:119: TypeError
```
2. we are forcing a deallocation in `krylov_exp` when torch need that tensor for its computational graph. This triggers
```python
E           RuntimeError: Cannot deallocate tensor

emu_base/utils.py:26: RuntimeError
```

### To decide
- [ ] The new test introduced to show the issue in `test_sv_backend.py` no longer make sense because we propose here to remove the logic.

### Deallocation
tensors cannot be deallocated inside an `autograd.Function`, even if they aren't in the computational graph
```python
class DoStuff(torch.autograd.Function):
    @staticmethod
    def forward(ctx, x: torch.Tensor):
        n = x.norm()
        print(x._use_count())
        print(x._base)
        deallocate_tensor(x)
        return n

    @staticmethod
    def backward(ctx, grad_output):
        return grad_output

t = torch.ones(10, 10)
print(t._use_count())
res = DoStuff.apply(t)
```
This cannot deallocate because the use count is 3. It is probably one for the object itself, one for the function call and one additional for torch stuff.

### Easy benchmark

Changes tries to fix the issue while preserving the memory improvements.

before changes, emu_sv getting_started_notebook
```
step = 1/600, RSS = 13.275 MB, Δt = 0.009 s
step = 2/600, RSS = 13.275 MB, Δt = 0.007 s
step = 3/600, RSS = 13.275 MB, Δt = 0.007 s
step = 4/600, RSS = 13.799 MB, Δt = 0.007 s
step = 5/600, RSS = 13.799 MB, Δt = 0.007 s
step = 6/600, RSS = 13.799 MB, Δt = 0.008 s
step = 7/600, RSS = 13.799 MB, Δt = 0.008 s
step = 8/600, RSS = 13.799 MB, Δt = 0.007 s
step = 9/600, RSS = 13.799 MB, Δt = 0.007 s
step = 10/600, RSS = 13.799 MB, Δt = 0.007 s
```

only removing deallocation
```
step = 1/600, RSS = 13.766 MB, Δt = 0.315 s
step = 2/600, RSS = 13.766 MB, Δt = 0.007 s
step = 3/600, RSS = 13.766 MB, Δt = 0.007 s
step = 4/600, RSS = 14.290 MB, Δt = 0.008 s
step = 5/600, RSS = 14.290 MB, Δt = 0.008 s
step = 6/600, RSS = 14.290 MB, Δt = 0.007 s
step = 7/600, RSS = 14.290 MB, Δt = 0.008 s
step = 8/600, RSS = 14.290 MB, Δt = 0.008 s
step = 9/600, RSS = 14.290 MB, Δt = 0.008 s
```

in-place krylov exp
```
step = 1/600, RSS = 13.242 MB, Δt = 0.336 s
step = 2/600, RSS = 13.242 MB, Δt = 0.007 s
step = 3/600, RSS = 13.242 MB, Δt = 0.007 s
step = 4/600, RSS = 13.766 MB, Δt = 0.008 s
step = 5/600, RSS = 13.766 MB, Δt = 0.008 s
step = 6/600, RSS = 13.766 MB, Δt = 0.008 s
step = 7/600, RSS = 13.766 MB, Δt = 0.008 s
step = 8/600, RSS = 13.766 MB, Δt = 0.008 s
step = 9/600, RSS = 13.766 MB, Δt = 0.008 s
```

### DIFF
--- a/emu_base/math/krylov_exp.py
+++ b/emu_base/math/krylov_exp.py
@@ -56,7 +56,7 @@ def krylov_exp_impl(
         for k in range(k_start, j + 1):
             overlap = torch.tensordot(lanczos_vectors[k].conj(), w, dims=w.dim())
             T[k, j] = overlap
-            w = w - overlap * lanczos_vectors[k]
+            w -= overlap * lanczos_vectors[k]
 
         n2 = w.norm()
         T[j + 1, j] = n2
@@ -71,7 +71,7 @@ def krylov_exp_impl(
                 result=result, converged=True, happy_breakdown=True, iteration_count=j + 1
             )
 
-        w = w / n2
+        w /= n2
         lanczos_vectors.append(w)
 
         # Compute exponential of extended T matrix


a-quelle (approved):
lgtm

PR #131, opened 2025-06-04
[emu-sv] master eq solver: energy observable support

murogrande:
- in `RydbergLindbladian` class: refactor of `__matmul__` in order to define the `expect` function (the callback requires the `expect` function)
- energy observable will be measured using the `expect` function in the callback
- added a unit test for `expect`
- small tutorial with master equation solver
- delete the `energy_sv_imp` function in `custom_callback_implementations`, it is not being used anywhere


test/emu_sv/test_end_to_end.py
```diff
@@ -160,8 +160,8 @@ def simulate_with_den_matrix(
             BitStrings(evaluation_times=times, num_shots=1000),
             Fidelity(evaluation_times=times, state=fidelity_state, tag_suffix="1"),
             Occupation(evaluation_times=times),
+            Energy(evaluation_times=times),
             # CorrelationMatrix(evaluation_times=times),
```
kbidzhiev, line 164: what about correlations ?

murogrande: I will remove the comment

kbidzhiev: one needs to add tests for correlations as well, you have occupation, energy, but not correlations

kbidzhiev (approved):

murogrande:
- First tutorial using density matrix and noisy sequence. 
- Adding a `expect` function for the `RydbergLindbladian` class.

PR #134, opened 2025-06-09
[emu-sv] density matrix: support for energy variance and second moment of energy

murogrande:
adding support for energy variance and second moment of energy observables
- update test regarding those observables

### DIFF
--- a/test/emu_sv/test_custom_callbacks.py
+++ b/test/emu_sv/test_custom_callbacks.py
@@ -121,7 +121,7 @@ def test_custom_energy_and_variance_and_second(noise) -> None:
     torch.manual_seed(1337)
 
     basis = ("r", "g")
-    num_qubits = 4
+    num_qubits = 4  # only even number for this test
     strings = {"rg" * int(num_qubits / 2): 1.0, "gr" * int(num_qubits / 2): 1.0}
     state = StateVector.from_state_amplitudes(eigenstates=basis, amplitudes=strings)
     config = SVConfig(gpu=gpu)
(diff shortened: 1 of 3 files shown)


emu_sv/custom_callback_implementations.py
```diff
@@ -147,3 +168,22 @@ def energy_second_moment_sv_impl(
...
+    config: EmulationConfig,
+    state: DensityMatrix,
+    hamiltonian: RydbergLindbladian,
+) -> torch.Tensor:
+    """
+    Custom implementation of the second moment of energy tr(ρH²) for the
+    lindblad equation solver.
+    """
+    h_dense_matrix = hamiltonian.h_eff(state.matrix)  # Hρ
+    gpu = False if (-state.matrix.get_device()) else True  # gpu = 0, cpu = -1
```
pablolh, line 185: I think something like this should work
`gpu = state.matrix.is_cuda`

murogrande: nice thanks

a-quelle (changes requested):
In `NoisyBackendImpl._evolve_step` it sets `self._current_H` to the Lindblad operator including the noise terms. This means that the energy and energy variance are computed on `h_eff` which is not hermitian, and the wrong values will be computed.

emu_sv/custom_callback_implementations.py
```diff
@@ -147,3 +168,22 @@ def energy_second_moment_sv_impl(
...
+    hamiltonian: RydbergLindbladian,
+) -> torch.Tensor:
+    """
+    Custom implementation of the second moment of energy tr(ρH²) for the
+    lindblad equation solver.
+    """
+    h_dense_matrix = hamiltonian.h_eff(state.matrix)  # Hρ
+    gpu = state.matrix.is_cuda
+
+    return hamiltonian.expect(
```
a-quelle, line 187: doing `h_dense_matrix.overlap(h_dense_matrix)` will be faster. It saves an application of `hamiltonian`

a-quelle: Mauro has alerted me to the fact that the default implementation of `EnergySecondMoment` and `EnergyVariance` in pulser are incorrect for density matrices. This makes the entire issue a bit moot. Let's leave the implementations as is. I will make a bugfix in pulser.

a-quelle (approved):
LGTM

PR #135, opened 2025-06-10
windows support

a-quelle:
Factored the max_RSS function into emu-base, and added some logic where on the cpu path on windows it just returns 0. rather than importing the non-existing resource package.


emu_base/utils.py
```diff
@@ -1,6 +1,11 @@
 from collections import Counter
 import random
 import torch
+import os
+
+unix_like = os.name != "nt"
```
elmerhej, line 6: nice!

elmerhej:
LGTM

Issue #143, opened 2025-06-20
State preparation errors


murogrande:
- support for density matrix and state vector

PR #150, opened 2025-07-11
[Emu-sv] support : state preparation error

murogrande:
- State preparation error is supported by `emu-sv` (State vector and density matrix)
- Changes in `emu-mps` should be done in order to move and adapt the state prep error logic implemented in emu-mps to `emu-base`
- added some tests
- refactor of callbacks in a way that now the callbacks decide which state to use `StateVector` or `DensityMatrix`

Related to #143


emu_sv/sv_config.py
```diff
@@ -126,47 +115,28 @@ def _expected_kwargs(self) -> set[str]:
...
             obs_copy = copy.deepcopy(obs)
 
             if isinstance(obs, Occupation):
                 obs_copy.apply = MethodType(  # type: ignore[method-assign]
-                    (
-                        qubit_occupation_sv_impl
-                        if use_state_vector
-                        else qubit_occupation_sv_den_mat_impl
-                    ),
+                    (choose_qubit_occupation_impl),
```
pablolh, line 124: parentheses are superfluous no?

emu_base/noise.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def pick_well_prepared_qubits(eta: float, n: int) -> torch.Tensor:
+    """
+    Randomly pick n booleans such that ℙ(True) = eta
+
+    Returns:
+        A BoolTensor of shape (n,) where each element is True with probability eta.
+    """
+    return torch.rand(n) < eta
```
pablolh, line 11: why did you invert the comparison?
`eta` is the probability of having a dark qubit, so this would instead be called `pick_dark_qubits` I think

emu_sv/custom_callback_implementations.py
```diff
@@ -63,6 +84,27 @@ def qubit_occupation_sv_den_mat_impl(
     return occupation.cpu()
 
 
+def choose_correlation_matrix_impl(
+    self: CorrelationMatrix,
+    *,
+    config: EmulationConfig,
```
pablolh, line 90: I suggest using `**kwargs` for `config` and `hamiltonian`, it feels better in that context because it is generic

emu_sv/custom_callback_implementations.py
```diff
@@ -117,6 +159,27 @@ def correlation_matrix_sv_den_mat_impl(
     return correlation.cpu()
 
 
+def choose_energy_variance_impl(
```
pablolh, line 162: to avoid repeating the same code, you could do something like

```
def choose(state_vector_version, density_matrix_version):
    def result(self, *, state: StateVector | DensityMatrix, **kwargs):
        if isinstance(state, StateVector):
            return state_vector_version(self, state=state, **kwargs)
        elif isinstance(state, DensityMatrix):
            return density_matrix_version(self, state=state, **kwargs)
        else:
            raise Exception("blabla")

    return result


choose_energy_variance_impl = choose(energy_variance_sv_impl, energy_variance_sv_den_mat_impl)

choose_correlation_matrix_impl = choose(correlation_matrix_sv_impl, correlation_matrix_sv_den_mat_impl)

# ...
```

emu_sv/custom_callback_implementations.py
```diff
@@ -16,6 +18,25 @@
 dtype = torch.float64
 
 
+def choose(
```
a-quelle-pasqal, line 21: this is a very nice pattern, but this function is not a callback. Shouldn't we put the implementation in `utils` or something?

PR #153, opened 2025-08-13
DMRGBackendImpl: Fixing edge cases and adding flag in config file

elmerhej:
Fixes in the DMRG routine:
- [x] Fix case when DMRG needs to update the first two and last two sites
- [x] Correct indexing during left bath update

Adding a flag in the config file
- [x] Add `adiabatic_evolution` flag to the config file to allow the user to enable the DMRG solver

Updating tests
- [x] Update mocktests to match the corrected DMRG behavior

### DIFF
--- a/emu_mps/mps_backend_impl.py
+++ b/emu_mps/mps_backend_impl.py
@@ -253,6 +253,7 @@ class MPSBackendImpl:
         initial_state.truncate()
         initial_state *= 1 / initial_state.norm()
         self.state = initial_state
+        self.state.orthogonalize(0)
 
     def init_hamiltonian(self) -> None:
         """
@@ -710,8 +711,6 @@ class DMRGBackendImpl(MPSBackendImpl):
                 f"you are using: {mps_config.noise_model.noise_types}"
             )
         super().__init__(mps_config, pulser_data)
-        self.init()
-        self.state.orthogonalize(0)
         self.previous_energy: Optional[float] = None
         self.current_energy: Optional[float] = None
         self.sweep_count: int = 0
--- a/test/emu_mps/test_mps_backend_impl.py
+++ b/test/emu_mps/test_mps_backend_impl.py
@@ -514,6 +514,7 @@ def test_left_to_right_update(
     mock_minimize.return_value = (torch.tensor([[1.0]]), torch.tensor([[2.0]]), 0.5)
 
     dmrg = create_dmrg_mock()
+    dmrg.init()
     dmrg.sweep_index = 1
     dmrg.swipe_direction = SwipeDirection.LEFT_TO_RIGHT
     dmrg.left_baths = [torch.zeros(1)]
@@ -554,6 +555,7 @@ def test_right_to_left_update(
     )
 
     dmrg = create_dmrg_mock()
+    dmrg.init()
     dmrg.sweep_index = 2
     dmrg.swipe_direction = SwipeDirection.RIGHT_TO_LEFT
     dmrg.left_baths = [torch.zeros(1, 1, 1, dtype=torch.complex128)] * 2
--- a/emu_mps/mps_config.py
+++ b/emu_mps/mps_config.py
@@ -56,7 +56,7 @@ class MPSConfig(EmulationConfig):
             therefore this interval is only a lower bound.
         solver: chooses the solver algorithm to run a sequence.
             Two options are currently available:
-            ``TDVP`` (default), which performs ordinary time evolution,
+            ``TDVP``, which performs ordinary time evolution,
             and ``DMRG``, which adiabatically follows the ground state
             of a given adiabatic pulse.
         kwargs: arguments that are passed to the base class


emu_mps/mps_backend_impl.py
```diff
@@ -702,6 +702,17 @@ def __init__(
         max_sweeps: int = 999,
         residual_tolerance: float = 1e-7,
     ):
+        if pulser_data.has_lindblad_noise is True:
```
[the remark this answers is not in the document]
elmerhej: Done

emu_mps/mps_backend_impl.py
```diff
@@ -747,21 +758,30 @@ def progress(self) -> None:
         self.current_energy = energy
 
         # updating baths and orthogonality center
-        if self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT:
+        if (
+            self.sweep_index < self.qubit_count - 2
+            and self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT
+        ):
```
kbidzhiev, line 764: in this set of if statements I suggest to put first direction and after that comparison of sweep indices.

The first part of the comparison is going to be evaluated after checking the direction. AND it's going to help shrink the code, i.e treat the edge case as a subcase of direction

emu_mps/mps_backend_impl.py
```diff
@@ -747,21 +758,30 @@ def progress(self) -> None:
...
+                    self.hamiltonian.factors[left_idx],
                 ).to(self.state.factors[right_idx].device)
             )
             self.right_baths.pop()
             self.sweep_index += 1
 
-            if self.sweep_index == self.qubit_count - 1:
+            if (
+                self.sweep_index == self.qubit_count - 2
+                and self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT
```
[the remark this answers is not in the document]
elmerhej: I restructured the code, as we previously discussed

emu_mps/mps_backend_impl.py
elmerhej: Similar to above
Done

emu_mps/mps_backend_impl.py
```diff
@@ -728,63 +739,72 @@ def progress(self) -> None:
...
+                    self.state.factors[idx],
+                    self.hamiltonian.factors[idx],
+                ).to(self.state.factors[idx + 1].device)
             )
             self.right_baths.pop()
             self.sweep_index += 1
 
-            if self.sweep_index == self.qubit_count - 1:
-                self.swipe_direction = SwipeDirection.RIGHT_TO_LEFT
+        if self.sweep_index == self.qubit_count - 2:
```
kbidzhiev, line 777: `if idx < self.qubit_count - 2:`

emu_mps/mps_backend_impl.py
```diff
@@ -728,63 +739,72 @@ def progress(self) -> None:
...
+            ham_factors=self.hamiltonian.factors[idx : idx + 2],
             baths=(self.left_baths[-1], self.right_baths[-1]),
             orth_center_right=orth_center_right,
             config=self.config,
             residual_tolerance=self.residual_tolerance,
         )
-        self.state.factors[left_idx], self.state.factors[right_idx] = new_L, new_R
-        self.state.orthogonality_center = right_idx if orth_center_right else left_idx
+        self.state.factors[idx], self.state.factors[idx + 1] = new_L, new_R
+        self.state.orthogonality_center = idx + 1 if orth_center_right else idx
```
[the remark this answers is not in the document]
elmerhej: I left it outside to avoid repetitive code patterns, but can definitely be done

emu_mps/mps_backend_impl.py
kbidzhiev: I've updated my comment. not it is with "suggestion" mark

emu_mps/mps_backend_impl.py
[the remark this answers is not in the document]
kbidzhiev: I see it better to move to iterator update logic because it is a iteration variable and, notice how you defined it 
 so it's clearly related to direction, not position

```
orth_center_right = self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT
self.state.orthogonality_center = idx + 1 if orth_center_right else idx
```
 is actually

```
orth_center_right = self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT
self.state.orthogonality_center = idx + 1 if SwipeDirection.LEFT_TO_RIGHT else idx
```

emu_mps/mps_backend_impl.py
```diff
@@ -728,63 +739,72 @@ def progress(self) -> None:
...
+            self._left_to_right_update(idx)
+        elif self.swipe_direction == SwipeDirection.RIGHT_TO_LEFT:
+            self._right_to_left_update(idx)
+        else:
+            raise Exception("Did not expect this")
+
+        self.save_simulation()
+
+    def _left_to_right_update(self, idx: int) -> None:
+        if self.sweep_index < self.qubit_count - 2:
```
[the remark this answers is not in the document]
elmerhej: Done

kbidzhiev (approved):

emu_mps/mps_backend_impl.py
```diff
@@ -728,63 +735,72 @@ def progress(self) -> None:
...
+        self.state.orthogonality_center = idx + 1 if orth_center_right else idx
         self.current_energy = energy
 
         # updating baths and orthogonality center
         if self.swipe_direction == SwipeDirection.LEFT_TO_RIGHT:
+            self._left_to_right_update(idx)
+        elif self.swipe_direction == SwipeDirection.RIGHT_TO_LEFT:
+            self._right_to_left_update(idx)
+        else:
+            raise Exception("Did not expect this")
```
kbidzhiev, line 757: This part is a bit suspicious, I think it can be removed. Above you have an assert, which does the job 

```
        assert self.swipe_direction in (
            SwipeDirection.LEFT_TO_RIGHT,
            SwipeDirection.RIGHT_TO_LEFT,
        ), "Unknown Swipe direction"
```

test/emu_mps/test_mps_backend_impl.py
```diff
@@ -491,8 +492,76 @@ def test_progress_at_right_mps_boundary(
...
+    assert len(dmrg.right_baths) == 1
+    assert dmrg.swipe_direction == SwipeDirection.RIGHT_TO_LEFT
+
+
+@patch("emu_mps.mps_backend_impl.new_left_bath")
+@patch("emu_mps.mps_backend_impl.minimize_energy_pair")
+@patch("emu_mps.mps_backend_impl.update_H")
+@patch("emu_mps.mps_backend_impl.make_H")
+@patch("emu_mps.mps_backend_impl.right_baths")
+def test_left_to_right_update(
```
a-quelle-pasqal, line 507: this test fails on my machine because it mixes gpu and cpu tensors. We can leave it as is for now, but you write an end-to-end test, test the sequence on the cluster when running it with `0,1,2` gpu's and see if any errors come out. It could just be the test, but there are some subtleties with moving tensors around that cannot be tested in the pipeline.

PR #156, opened 2025-09-17
end-to-end tests for dmrg

elmerhej:
writing end-to-end tests for DMRG

### DIFF
--- a/test/emu_mps/test_endtoend.py
+++ b/test/emu_mps/test_endtoend.py
@@ -28,6 +28,7 @@ import pulser.noise_model
 from pulser.backend import Results
 
 from emu_mps.mps_backend_impl import MPSBackendImpl
+from emu_mps.solver import Solver
 from emu_mps.solver_utils import right_baths
 from test.utils_testing import (
     pulser_afm_sequence_grid,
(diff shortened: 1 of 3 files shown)


test/emu_mps/test_endtoend.py
```diff
@@ -342,6 +345,147 @@ def test_end_to_end_afm_ring() -> None:
...
+    state_fin = result.state[final_time]
+    fidelity_fin = result.fidelity_1[final_time]
+    max_bond_dim = state_fin.get_max_bond_dim()
+    fidelity_st = create_antiferromagnetic_mps(num_qubits)
+    occupation_even_sites = occupation[0::2]
+    occupation_odd_sites = occupation[1::2]
+
+    assert max_bond_dim == 4
+    # check that the output state is the AFM state
+    assert fidelity_st.overlap(state_fin) == approx(fidelity_fin, abs=1e-10)
```
a-quelle-pasqal, line 379: For all three tests, I think it's a good idea to have the expected final fidelity in the test as a hardcoded number. This will cause the test to fail if the time evolution changes somehow. It's true that for the bitstring samples this will probably also occur, but those will also change if you reorder calls to the RNG, while the final fidelity will not.
The assert in this line, while totally fine, mostly tests the consistency of the `Fidelity` observable.

a-quelle-pasqal (approved):
all good

Issue #161, opened 2025-09-22
[emu-mps] leakage: mps supporting 3 level systems


murogrande:
### Make changes in emu-mps for supporting leakage

In order to support leakage, I will start from the state (MPS), converting it into a 3 level system. 

I would like to make changes until the 3 level system is supported and at the same the 2 level system. 

### End of issue

Previous tests (2 level systems) should not be changed and the new tests (for 3 level systems) should follow the same structure and should be adapted for 3 level systems.

PR #170, opened 2025-10-14
Noise: support for leakage in MPS 

murogrande:
Related to #161, 
The leakage will be our 3rd, basis $|0\rangle = [0,0,1]$. In this MR, I am modifying the `mps.py` file and also the `test_mps.py` in order to support 3 level systems. 
- The new basis is: ${"r","g", "x"}$.
- When the leakage is activated, the `sample` will threat this as a "0". I add tests in `test_probablitysample.py` where you will see this behavior. 

NOTE: the correlation matrix and the occupation operators are not supported yet. Thus, in `test_mps.py` I am not modifying these. I will do it in the next MR where I will update the MPO class in order to support 3 level systems

### DIFF
--- a/emu_mps/mps.py
+++ b/emu_mps/mps.py
@@ -265,7 +265,7 @@ class MPS(State[complex, torch.Tensor]):
 
                 batch_outcomes[:, qubit] = outcomes
 
-                # (batch_size, bond_dim)
+                # expected shape (batch_size, bond_dim)
                 batched_accumulator = batched_accumulator[rangebatch, outcomes, :]
 
             shots_done += batch_size


emu_mps/mps.py
```diff
@@ -104,11 +110,22 @@ def make(
...
+
+        elif len(eigenstates) == 3:  # (g,r,x)
+            ground_state = [
+                torch.tensor([[[1.0], [0.0], [0.0]]], dtype=dtype)
                 for _ in range(num_sites)
-            ],
+            ]
+
+        else:
+            raise ValueError("Unsupported qudit provided. Upto 3 levels supported.")
```
a-quelle-pasqal: what about digital, that has something like g and h no?

test/emu_mps/test_mps.py
```diff
@@ -9,6 +9,15 @@
 
 from test.utils_testing import ghz_state_factors
 
+dtype = torch.complex128
+device = "cpu"
```
a-quelle-pasqal: is this used?

emu_mps/mps.py
```diff
@@ -67,6 +71,7 @@ def __init__(
 
         self.factors = factors
         self.num_sites = len(factors)
+        self.dim = len(self.eigenstates)
```
a-quelle-pasqal: I think we should assert that the physical dimension of all the state tensors is equal to this. When you add this, you will notice that some tests start failing.

emu_mps/mps.py
murogrande: but we are not supporting the `digital` basis in emulators but I see your point. Let me specify this:
`Unsupported basis provided. The supported basis are: `{(0,1),("r","g"),("r","g","x")} `
What do you think about this comment?

test/emu_mps/test_endtoend.py
```diff
@@ -328,8 +328,8 @@ def test_end_to_end_afm_ring() -> None:
     assert max_bond_dim == 29
     assert fidelity_st.overlap(state_fin) == approx(fidelity_fin, abs=1e-10)
 
-    assert bitstrings["1010101010"] == 129
-    assert bitstrings["0101010101"] == 135
+    assert bitstrings["1010101010"] == 138
```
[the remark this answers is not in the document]
murogrande: I update now all the failing tests are patched with the `cpu_multinomial_wrapper`

emu_mps/mps.py
a-quelle-pasqal: yes, that's better

emu_mps/custom_callback_implementations.py
```diff
@@ -25,9 +25,18 @@ def qubit_occupation_mps_impl(
...
-    op = torch.tensor(
-        [[[0.0, 0.0], [0.0, 1.0]]], dtype=torch.complex128, device=state.factors[0].device
-    )
+    if state.dim == 2:
+        op = torch.tensor(
+            [[[0.0, 0.0], [0.0, 1.0]]],
+            dtype=torch.complex128,
+            device=state.factors[0].device,
+        )
+    elif state.dim == 3:
```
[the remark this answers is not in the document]
a-quelle-pasqal: that is a good point, you could create a `torch.zeros(state.dim,state.dim)` and then just put the `1,1` element to `1` it would collapse all the if-else logic into those two lines.

emu_mps/mps.py
```diff
@@ -182,7 +201,7 @@ def get_max_bond_dim(self) -> int:
         Returns:
             the largest bond dimension in the state
         """
-        return max((x.shape[2] for x in self.factors), default=0)
+        return max((factor.shape[2] for factor in self.factors), default=0)
```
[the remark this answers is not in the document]
a-quelle-pasqal: I've also seen situations where the bond-dimension oscillates a bit, so on average it grows towards the inside, but then the odd sites have slightly higher entanglement than the even or vice versa. So in general, there is no way of knowing at what site the bond dimension peaks.

emu_mps/mps.py
```diff
@@ -40,18 +41,21 @@ def __init__(
         eigenstates: Sequence[Eigenstate] = ("r", "g"),
     ):
         """
-        This constructor creates a MPS directly from a list of tensors. It is for internal use only.
+        This constructor creates a MPS directly from a list of tensors. It i
```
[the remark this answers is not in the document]
murogrande: good Elie !

emu_base/pulser_adapter.py
```diff
@@ -84,13 +84,16 @@ def _extract_omega_delta_phi(
     elif "XY" in sequence_dict and len(sequence_dict) == 1:
         locals_a_d_p = sequence_dict["XY"]
     else:
-        raise ValueError("Only `ground-rydberg` and `mw_global` channels are supported.")
+        raise ValueError(
+            "Only `ground-rydberg` and `mw_global` channels " "are supported."
```
kbidzhiev: channels " "are -> channels are

test/emu_mps/test_probabilitysample.py
```diff
@@ -1,30 +1,49 @@
...
+
 dtype = torch.complex128
 
 
-def test_sampling_ghz5_mps():
-    device = "cpu"
+@pytest.mark.parametrize(
+    "basis",
+    (
+        ("0", "1"),
```
kbidzhiev: just (r, g) as well to check the full space ? though the sampling part is expected do to the same

test/emu_mps/test_probabilitysample.py
```diff
@@ -37,7 +56,94 @@ def test_not_orthogonalized_state():
...
+
+
+def test_with_leakage():
+    torch.manual_seed(seed)
+    shots = 1000
+    half_atoms = 4  # half the number of qubits
+
+    ket1 = torch.tensor([[[0], [1], [0]]], dtype=torch.complex128, device=device)  # |r>
+    ket0 = torch.tensor([[[1], [0], [0]]], dtype=torch.complex128, device=device)  # |g>
+    wall_state = [ket1] * half_atoms + [ket0] * half_atoms  # wall state in rydberg basis
```
kbidzhiev: correct name is "domain wall"

test/utils_testing/utils_testing.py
```diff
@@ -6,44 +6,120 @@
 import pulser
 import math
 
+# for testing purposes, reference to the real multinomial
+_real_multinomial = torch.multinomial
+
+
+def cpu_multinomial_wrapper(probs: torch.Tensor, num_samples: int, replacement=False):
+    """
+    For independent device *(cpu or gpu) tests. This is a function that
```
kbidzhiev: what * means ?

test/utils_testing/utils_testing.py
```diff
@@ -6,44 +6,120 @@
...
-        / torch.sqrt(torch.tensor([2.0], device=device, dtype=dtype))
-        * torch.tensor(
+    if dim == 2:
+        core_1 = (
+            1
+            / torch.sqrt(torch.tensor([2.0], device=device, dtype=dtype))
+            * torch.tensor(
+                [
+                    [
+                        [1.0 + 0.0j, 0.0 + 0.0j],
```
kbidzhiev: you can safely remove `0.0j` and make 1.0 to 1, because type you use is complex

test/emu_mps/test_probabilitysample.py
murogrande: Thanks for the clarification

test/utils_testing/utils_testing.py
murogrande: the * is an error. Good catch

emu_mps/mps.py
```diff
@@ -206,64 +240,52 @@ def sample(
...
         while shots_done < num_shots:
             batch_size = min(max_batch_size, num_shots - shots_done)
             batched_accumulator = torch.ones(
-                batch_size, 1, dtype=torch.complex128, device=self.factors[0].device
+                batch_size, 1, dtype=dtype, device=self.factors[0].device
             )
 
-            batch_outcomes = torch.empty(batch_size, self.num_sites, dtype=torch.bool)
-
+            batch_outcomes = torch.empty(batch_size, self.num_sites, dtype=torch.long)
```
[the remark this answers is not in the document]
kbidzhiev: this is a datatype, I don't see any reason why it can cause runtime error. You might say that the number of samples can be large, but default `int` is large enough. I don't have anything against `long`. It just looks out of turn and unnecessary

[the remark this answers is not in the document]
murogrande: I will change it

test/emu_mps/test_probabilitysample.py
```diff
@@ -37,7 +57,96 @@ def test_not_orthogonalized_state():
...
+    assert bitstrings.get("000") == 501
+
+
+def test_sampling_wall():
+    torch.manual_seed(seed)
+    shots = 1000
+    half_atoms = 4  # half the number of qubits
+
+    factor1 = torch.tensor([[[0], [1]]], dtype=torch.complex128, device=device)
+    factor2 = torch.tensor([[[1.0], [0.0]]], dtype=torch.complex128, device=device)
```
kbidzhiev: the same 1.0 can be 1 and 0.0 can be 0

test/utils_testing/utils_testing.py
```diff
@@ -6,44 +6,120 @@
...
-        / torch.sqrt(torch.tensor([2.0], device=device, dtype=dtype))
-        * torch.tensor(
+    if dim == 2:
+        core_1 = (
+            1
+            / torch.sqrt(torch.tensor([2.0], device=device, dtype=dtype))
+            * torch.tensor(
+                [
+                    [
+                        [1.0, 0.0],
```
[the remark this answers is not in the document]
kbidzhiev: okaaay :)

PR #171, opened 2025-10-14
Em/documentation dmrg

elmerhej:
Updating current documentation with the new DMRG solver.

- Changed the name of `TDVP.md` to `algorithms.md` where both `TDVP` and `DMRG` are described.
- Described DMRG with some level of detail, showing why it can be useful in the context of the `emulators` package. 
- updated the config docs and other files in the `advanced` folder to shed some light on both solvers TDVP and DMRG.

### DIFF
--- a/docs/emu_mps/advanced/config.md
+++ b/docs/emu_mps/advanced/config.md
@@ -56,7 +56,7 @@ num_gpus_to_use = 2  # use 2 GPUs if available, otherwise fallback to 1 or CPU
 ## optimize_qubit_ordering
 The `optimize_qubit_ordering` parameter enables the reordering of qubits in the register. This can be useful in cases where the initial qubit ordering (chosen by the user) is not optimal. In such cases, setting `optimize_qubit_ordering = True` re-orders the qubits more efficiently, and that has been shown to improve performance and accuracy. The default value is `False`.
 
-**Note:** enabling this option disables certain features.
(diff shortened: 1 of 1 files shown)


a-quelle-pasqal (changes requested):
The resource_estimation.md file contains much that is also true for DMRG. Currently it only mentions TDVP in many places, and that could be relaxed a bit.

docs/emu_mps/advanced/config.md
```diff
@@ -32,10 +36,36 @@ In addition to the above procedure, at each truncation step, no more than `max_b
...
+Setting `num_gpus_to_use = 0` runs the entire computation on the CPU.
+Using multiple GPUs can reduce memory usage per GPU, though the overall runtime remains similar.
+
+**Example:**
+num_gpus_to_use = 2  # use 2 GPUs if available, otherwise fallback to 1 or CPU
+
+## optimize_qubit_ordering
+The `optimize_qubit_ordering` parameter enables the reordering of qubits in the register. This can be useful in cases where the initial qubit ordering (chosen by the user) is not optimal. In such cases, setting `optimize_qubit_ordering = True` re-orders the qubits more efficiently, and that has been shown to improve performance and accuracy. The default value is `False`.
+
+**Note:** enabling this option disables certain features.
```
kbidzhiev: Could you say "Enabling this option is not compatible with certain features, such as reading a user-provided initial state."

Issue #181, opened 2025-10-30
[emu-base /pulser basis] basis inconsistency in the XY basis


murogrande:
Pulser and emu-* agree in the XY basis convention. However, in the following lines of[ code](https://github.com/pasqal-io/emulators/blob/main/emu_base/jump_lindblad_operators.py#L41-L44), we do not need `torch.flip` in the XY basis. 
Thus, this is an error.

PR #183, opened 2025-11-04
[Bug] Jump operators that come from pulser need (or not) to be arranged depending the basis 

murogrande:
Related #181 
Lindblad operators are coming from pulser. Now, Pulser supports multiple bases and the bases are written in different order. 
For instance, in the case of Rydberg basis pulser uses the following:
$|r\rangle=[1.0,0.0]$ and $|g\rangle=[0.0,1.0]$.

In the case of XY, pulser uses the following:
$|0\rangle=[1.0,0.0]$ and $|1\rangle=[0.0,1.0]$.

For more information, check this pulser [link](https://pulser.readthedocs.io/en/stable/conventions.html).

In our case, emu-mps (and emu-sv) always uses the same basis :
$|0\rangle = |g\rangle= [1.0,0.0]$ and $|1\rangle= |g\rangle=[0.0,1.0]$.

After this, if a user is using the Rydberg basis and wants to feed a jump operator to emu-mps or emu -sv, we are rearranging the elements of the jump operator according to our basis. 
However, if the user wants to feed a jump operator using the XY basis, emu-mps (and emu-sv) should not do anything to the elements of the jump operator but internally, emu-mps (and emu-sv) is rearranging the elements in this case, also.

This discrepancy between pulser and emu-mps regarding Lindblad operators is being addressed in this MR

### DIFF
--- a/test/emu_base/test_adapter.py
+++ b/test/emu_base/test_adapter.py
@@ -736,7 +736,7 @@ def test_parsed_sequence(mock_data):
     )
 
     cutoff_interaction_matrix = torch.tensor(
-        [[0, 0, -0.4], [0, 0, 0], [-0.4, 0, 0]],
+        [[0.0, 0.0, -0.4], [0.0, 0.0, 0.0], [-0.4, 0.0, 0.0]],
         dtype=torch.float64,
     )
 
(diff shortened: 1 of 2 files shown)


murogrande:
I notice that the `NoiseModel` in Pulser accepts Numpy array (and torch tensors and Qutip operators) as jump operators. But emu-mps (and emu-sv) only supports `torch.Tensor` and it will throw an error in the other cases.

a-quelle-pasqal (approved):
LGTM

Issue #149, opened 2025-07-09
[emu_mps] refactoring the MPSBackendImpl


elmerhej:
As discussed in the following PR: https://github.com/pasqal-io/emulators/pull/146#pullrequestreview-2997645913
the MPSBackendImpl needs refactoring as such:

- write a base class `MPSBackendImplBase` that includes all common methods for `TDVP` and `DMRG`. this class should have `progress` and `sweep_complete` as abstract methods
- write a subclass: TDVPBackendImpl with its own progress/sweep_complete methods
- write a subclass: DMRGBackendImpl with its own progress/sweep_complete methods
- `NoisyMPSBackendImpl` should become `NoisyTDVPBackendImpl`, thereby inheriting all methods from `TDVPBackendImpl` and overriding `sweep_complete`

PR #184, opened 2025-11-06
Y padawan/149 refactoring mpsbackendimpl

YPadawan:
This PR is being opened in order to solve issue #149. Several of the required refactoring has been done. But there are still some doubts and there are still some details to check before eventually merging.

- [x] `MPSBackendImplBase` is now an abstract base class where `sweep_complete` and `progress` are abstract methods
- [x] A `TDVPBackendImpl` has been implemented with the sweep_complete and progress methods that were previously in `MPSBackendImplBase`. Comments indicated that it was TDVP, but it needs to be confirmed
- [x] `NoisyMPSBackendImpl` has become `NoisyTDVPBackendImpl`, inheriting from `TDVPBackendImpl` and overriding sweep_complete

In addition to this a new module named **permutators.py** has been created where all functions making permutations were transferred for clarity purposes. Only backend classes remain in **mps_backend_impl.py**. Maybe more refactoring should be considered.

- [x] Linting has been made in **mps_backend_impl.py** and **permutators.py** using pylint.

Documentation is still lacking, but can be done before merging.

Tests of **test_mps_backend_impl.py** are running and all of them pass. But a change has been made in L307 of this file, because the two tensors that were being compared were located in different places, one was in the cpu the second in the gpu. In L307, the relevant tensor is explicitly sent to the cpu (for now). This might be a local problem, or it might have something to do with the test itself. It needs investigating.

When testing all other test files, apart from those that took too long to run almost all of them passed for emu_mps. There was one failure that should be mentioned in here:
*test_solver_utils.py::test_evolve_pair*
the torch.allclose failed. (It should be out of scope of the issue).


YPadawan:
Another question, I have opened the PR to merge with the main branch, maybe there is another one where merging usually happens ? Before merging with the main branch ?

PR #191, opened 2025-11-25
State normalization tolerance is not compatible with pulser

a-quelle-pasqal:
The `_from_state_amplitudes` functions in `MPS` and `StateVector` normalize the state with a tolerance of `1e-5`. This is much higher than the `1e-12` tolerance that pulser uses in checks for whether the state has been changed after creation (in which case serialization is not supported).

I'm changing the tolerance to comply with that in pulser. The following script fails on the old code, and now passes, when using either emu-sv or emu-mps objects:

```
import numpy as np

from pulser import Register, Pulse, Sequence
from pulser.devices import MockDevice
from emu_sv import SVBackend, SVConfig, StateVector
from emu_mps import MPS, MPSBackend, MPSConfig
import copy

natoms = 2
reg = Register.rectangle(1,natoms, spacing=8.0,prefix="q")
import torch
torch.set_printoptions(precision=16)
seq = Sequence(reg, MockDevice)
seq.declare_channel("ch0", "rydberg_global")
duration = 500
pulse = Pulse.ConstantPulse(duration, 4 * np.pi, 0.0, 0.0)
seq.add(pulse, "ch0")
# amp_full = {'g': (0.7071067811865476+0j), 'r': (0.7071067811865475+0j)} # where  the error was found
amp_full = {'g'*natoms: (0.7071+0.0j), 'r'*natoms: (0.7071+0.0j)}

#amp_full_but_one = {'r': (0.7071067811865475+0j)}  # this is working


state = MPS.from_state_amplitudes(
    eigenstates=("r","g"),
    amplitudes=amp_full
)
config = MPSConfig(initial_state = copy.deepcopy(state))
backend = MPSBackend(sequence=seq, config=config)
result = backend.run()

reconstructed = MPS.from_state_amplitudes(
    eigenstates=("r","g"),
    amplitudes=state._amplitudes  
)

#checking with pulser
print(abs(float(state.overlap(reconstructed)) - 1.0))

config.to_abstract_repr()
```

### DIFF
--- a/test/emu_mps/test_mps_config.py
+++ b/test/emu_mps/test_mps_config.py
@@ -124,7 +124,6 @@ def test_serialization_with_state():
     natoms = 2
     reg = pulser.Register.rectangle(1, natoms, spacing=8.0, prefix="q")
 
-    torch.set_printoptions(precision=16)
     seq = pulser.Sequence(reg, pulser.MockDevice)
     seq.declare_channel("ch0", "rydberg_global")
     duration = 500
--- a/test/emu_sv/test_svconfig.py
+++ b/test/emu_sv/test_svconfig.py
@@ -82,9 +82,7 @@ def test_expected_kwargs() -> None:
 def test_serialization_with_state():
     natoms = 2
     reg = pulser.Register.rectangle(1, natoms, spacing=8.0, prefix="q")
-    import torch
 
-    torch.set_printoptions(precision=16)
     seq = pulser.Sequence(reg, pulser.MockDevice)
     seq.declare_channel("ch0", "rydberg_global")
     duration = 500


emu_sv/state_vector.py
```diff
@@ -66,7 +66,8 @@ def _normalize(self) -> None:
 
         norm = torch.linalg.vector_norm(self.vector)
 
-        if not torch.allclose(norm, torch.ones_like(norm)):
+        # This must duplicate the tolerance in pulsers State._to_abstract_repr
+        if abs(self.norm() ** 4 - 1.0) > 1e-12:
```
elmerhej: why are you raising the norm (square root of <pis|psi>) to the power of 4? why not just 2?

emu_sv/state_vector.py
```diff
@@ -268,7 +269,7 @@ def _from_state_amplitudes(
             bin_to_int = int(
                 state.replace(one, "1").replace("g", "0"), 2
             )  # "0" basis is already in "0"
-            accum_state.vector[bin_to_int] = torch.tensor([amplitude])
+            accum_state.vector[bin_to_int] = amplitude  # type: ignore [assignment]
```
murogrande: if we just add the dtype?
`accum_state.vector[bin_to_int] = torch.tensor([amplitude],dtype=dtype)`

and you delete the `type:ignore`

murogrande (approved):
LGTM!

PR #197, opened 2025-12-02
Em/fix pulser adapter bug

elmerhej:
## Fix `pulser adapter` in `emu-base`: how the code should handle qubits that are not subjected to a pulse:

- Updated `_extract_omega_delta_phi` in `emu_base` to assign zero amplitude, detuning, and phase for qubits that don't feel the pulser sequence.
- Added a unit test `test_extract_omega_delta_phi_missing_qubit` to test the behaviour:

### DIFF
--- a/test/emu_base/test_adapter.py
+++ b/test/emu_base/test_adapter.py
@@ -925,94 +925,12 @@ def test_extract_omega_delta_phi_missing_qubit():
         qubit_ids=qubit_ids,
         target_times=target_times,
     )
-
     # Check values of omega for q0 and q2
-    assert (
-        omega[0, 0]
-        == (
-            mock_pulser_dict["ground-rydberg"]["q0"]["amp"][0]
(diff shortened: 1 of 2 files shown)


test/emu_base/test_adapter.py
```diff
@@ -890,3 +890,47 @@ def test_non_lindbladian_noise():
...
+        target_times=target_times,
+    )
+
+    # q1 should have zero amplitude, detuning and phase
+    assert torch.all(omega[:, 1] == 0)
+    assert torch.all(delta[:, 1] == 0)
+    assert torch.all(phi[:, 1] == 0)
+
+    # q0 and q2 should have a non-zero value of their amps
+    assert omega[:, 0][0] == 1.5  # due to linear interpolation formula
```
murogrande: here I will test all the values of omega, not the first one only. You can give by hand the numbers.

PR #202, opened 2025-12-19
Mm/xy rydberg 3 level

murogrande:
READY BUT THIS NEED TO WAIT
XY-Rydebrg interaction, it is working but it will live in this branch (in the fridge) until the time comes.


a-quelle-pasqal:
because of #248 I think we can close this now.

PR #205, opened 2026-01-02
Kb/fractional dt

### LINKED ISSUE #169: evaluation_times inconsistencies between backends
All backends should return the observables at the times required by the user.

### Description

As per goal above, when setting some `evaluation_times` in an observable, i.e. `BitStrings(evaluation_times=evalulation_times)` all backends should return a list of evaluation of that observable at the required times. In particular:

- [x] `emu_sv/mps` do not return the initial value observable (at time 0), while `pulser_simulation` yes
- [x] evaluation times are rounded differently between `emu_sv/mps` and `pulser_simulation`
- [ ] `pulser_simulation` does not accept numpy evaluation times. It would be desirable that all backends accept both list or numpy array as evaluation times (~ArrayLike objects)
- [x] check serialization of `atom_order` in `Results`. emu-sv seemingly does not store the atom_order.

### To reproduce:
```python
import logging
import numpy as np
from pulser import Sequence, Register, Pulse, AnalogDevice
from pulser.backend import BitStrings
from emu_sv import SVBackend, SVConfig
from pulser_simulation import QutipBackendV2, QutipConfig

reg = Register({"q0":[-3,0], "q1":[3,0]})
seq = Sequence(reg, AnalogDevice)
seq.declare_channel("ryd", "rydberg_global")
pulse = Pulse.ConstantPulse(400,1,0,0)
seq.add(pulse, channel="ryd")

evalulation_times = np.linspace(0,1,13).tolist()
bitstrings = BitStrings(evaluation_times=evalulation_times)
observables = (bitstrings,)

sv_config = SVConfig(observables = observables, log_level=logging.WARN)
sv_backend = SVBackend(seq,config=sv_config)
sv_results = sv_backend.run()

qutip_config = QutipConfig(observables = observables)
qutip_backend = QutipBackendV2(seq,config=qutip_config)
qutip_results = qutip_backend.run()


sv_times = sv_results.get_result_times(bitstrings)
qutip_times = qutip_results.get_result_times(bitstrings)

# I fail because emu_sv is missing time = 0 result
assert np.allclose(sv_times,qutip_times)

# I fail because times are slightly different
assert np.allclose(sv_times,qutip_times[1:])
```

kbidzhiev:
- [x] evaluation times are the same between `emu_sv/mps` and `pulser_simulation`
- [x] check serialization of atom_order in Results

**Summary**

PR generalises observable times from `int` to `float` to match `Pulser` behavior.

**Details**

Updates time generation/handling to support non-integer dt.

Adjusts related checks/comparisons to work reliably with floating-point times.

Partially closes #169.

### DIFF
--- a/emu_sv/sv_backend_impl.py
+++ b/emu_sv/sv_backend_impl.py
@@ -188,17 +188,17 @@ class BaseSVBackendImpl:
 
     def _save_statistics(self, step_idx: int) -> None:
         norm_time = self.target_times[step_idx] / self.target_times[-1]
-        if self._is_evaluation_time(self.statistics, norm_time):
-            self.statistics.data.append(time.time() - self.time)
-            self.statistics(
-                self._config,
-                norm_time,
-                self.state,
-                self._current_H,  # type: ignore[arg-type]
-                self.results,
-            )
-            self.time = time.time()
-            self._current_H = None
+
+        self.statistics.data.append(time.time() - self.time)
+        self.statistics(
+            self._config,
+            norm_time,
+            self.state,
+            self._current_H,  # type: ignore[arg-type]
+            self.results,
+        )
+        self.time = time.time()
+        self._current_H = None
 
     def _run(self) -> Results:
         self._apply_observables(0)  # at t == 0 for pulser compatibility


emu_base/pulser_adapter.py
```diff
@@ -39,29 +39,42 @@ def _get_all_lindblad_noise_operators(
...
     sequence_duration = sequence.get_duration(include_fall_time=config.with_modulation)
 
-    observable_times = set(range(0, sequence_duration, dt))
+    observable_times = set()
+
+    t = 0.0
+    while t < sequence_duration:
+        observable_times.add(round(t, ROUND_DIGITS))
+        t += dt
     observable_times.add(sequence_duration)
```
sgrava: This leaves out the final time, right? at which point it is added?

test/emu_base/test_adapter.py
```diff
@@ -837,6 +837,7 @@ def test_get_target_times_with_obs_eval_time(with_modulation):
     duration = 123
     dt = 3
 
+    eval_times = [0.5, 0.9]
```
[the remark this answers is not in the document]
kbidzhiev: I'll thingk how to make this more explicit. The point is 123 * 0.5 gives fractional absolute time

test/emu_mps/test_endtoend.py
```diff
@@ -1165,21 +1166,27 @@ def test_end_to_end_observable_time_as_in_pulser():
...
-    assert mps_results.get_result_times(bitstrings) == bitstrings_eval_times
-    assert mps_results.get_result_times(occup) == occupation_eval_times
+    qutip_config = QutipConfig(observables=obs)
+    qutip_backend = QutipBackendV2(seq, config=qutip_config)
+    qutip_results = qutip_backend.run()
+
+    mps_occ_t = mps_results.get_result_times(occ)
+    q_occ_t = qutip_results.get_result_times(occ)
+
+    for m, q in zip(mps_occ_t, q_occ_t):
```
a-quelle-pasqal: zip just zips up to the length of the shortest list. It's best to also assert that both lists have the same length.

emu_base/pulser_adapter.py
```diff
@@ -39,29 +39,44 @@ def _get_all_lindblad_noise_operators(
...
+        10  # keeps 10 significant digits when rounding fractional observable_times
+    )
     sequence_duration = sequence.get_duration(include_fall_time=config.with_modulation)
 
-    observable_times = set(range(0, sequence_duration, dt))
+    observable_times: set[float] = set()
+
+    t = 0.0
+    while t < sequence_duration:
+        observable_times.add(round(t, ROUND_DIGITS))
```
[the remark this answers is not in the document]
kbidzhiev: I got rid of that concept

emu_sv/sv_backend_impl.py
```diff
@@ -82,7 +82,10 @@ def __init__(self, config: SVConfig, pulser_data: PulserData):
         self.full_interaction_matrix = pulser_data.full_interaction_matrix
         self.state: State
         self.time = time.time()
-        self.results = Results(atom_order=(), total_duration=self.target_times[-1])
+        self.results = Results(
+            atom_order=(),  # pulser_data.qubit_ids, TODO
```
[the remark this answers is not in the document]
kbidzhiev: done

test/emu_sv/test_end_to_end.py
```diff
@@ -859,24 +863,31 @@ def test_end_to_end_observable_time_as_in_pulser():
...
 
-    assert sv_results.get_result_times(bitstrings) == bitstrings_eval_times
-    assert sv_results.get_result_times(occup) == occupation_eval_times
+    sv_occ_t = sv_results.get_result_times(occ)
+    q_occ_t = qutip_results.get_result_times(occ)
+
+    assert np.allclose(sv_occ_t, q_occ_t), f"\nsv = {sv_occ_t}, \nq = {q_occ_t}"
+
+    # TODO
+    # doesn't work because of float access
```
a-quelle-pasqal: now that you use `np.allclose` do you still need the comments here and the for loop?

emu_mps/mps_backend_impl.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, mps_config: MPSConfig, pulser_data: PulserData):
             atom_order=optimat.permute_tuple(
                 pulser_data.qubit_ids, self.qubit_permutation
             ),
-            total_duration=self.target_times[-1],
+            total_duration=int(self.target_times[-1]),
```
murogrande: this is already an integer coming from `target_times`, what is the objective of having `int()` again?

emu_sv/sv_backend_impl.py
```diff
@@ -82,7 +82,10 @@ def __init__(self, config: SVConfig, pulser_data: PulserData):
         self.full_interaction_matrix = pulser_data.full_interaction_matrix
         self.state: State
         self.time = time.time()
-        self.results = Results(atom_order=(), total_duration=self.target_times[-1])
+        self.results = Results(
+            atom_order=pulser_data.qubit_ids,
+            total_duration=int(self.target_times[-1]),
```
murogrande: I have the feeling that `self.target_times[-1]` is not always `integer` and now I am wondering why? 
I see that you are asserting at the end of `def _get_target_times` function but overthought it is asserted before you are converting to `integer`

test/emu_mps/test_endtoend.py
```diff
@@ -1162,24 +1163,31 @@ def test_end_to_end_observable_time_as_in_pulser():
...
+    assert np.allclose(mps_occ_t, q_occ_t), f"\nmps = {mps_occ_t}, \nq = {q_occ_t}"
+
+    # TODO
+    # doesn't work because of float access
+    # mps_occ = mps_results.occupation[m]
+    # q_occ = qutip_results.occupation[q]
+    for mps_occ, q_occ in zip(mps_results.occupation, qutip_results.occupation):
+        assert np.allclose(
+            mps_occ, q_occ, atol=1e-4
+        ), f"mps_occ = {mps_occ}, q_occ = {q_occ}"
```
murogrande: I see bit strings were not taken into account, is it possible to add them? at least at the end of the sequence.

emu_mps/mps_backend_impl.py
murogrande: I see that you are assuming that `target_times array : list[float]` but in reality is not the case because the last element is integer. What happens if you delete the `int()` here in this function? do you have issues?

emu_sv/sv_backend_impl.py
kbidzhiev: I want developers to see that conversion is intended. Python is ok with removing `int()`.

I put it there because `target_times` must be `list[floats]` and `target_times[-1]` must be an `integer` to comply with pulser.  This information is implicit

murogrande (commented):
small comment

emu_sv/sv_backend_impl.py
```diff
@@ -143,9 +146,37 @@ def _compute_dt(self, step_idx: int) -> float:
     def _evolve_step(self, dt: float, step_idx: int) -> None:
         """One step evolution"""
 
+    def _is_evaluation_time(
+        self,
+        observable: Observable,
+        t: float,
+        tolerance: float = 1e-10,
```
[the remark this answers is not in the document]
kbidzhiev: Ok, if you are happy I'm resolving you comment

emu_base/pulser_adapter.py
```diff
@@ -38,30 +39,44 @@ def _get_all_lindblad_noise_operators(
...
-    sequence: pulser.Sequence, config: EmulationConfig, dt: int
-) -> list[int]:
-    sequence_duration = sequence.get_duration(include_fall_time=config.with_modulation)
-
-    observable_times = set(range(0, sequence_duration, dt))
-    observable_times.add(sequence_duration)
+def _unique_observable_times(config: EmulationConfig) -> set[float]:
+    """Collect unique evaluation times in [0, 1] for all observables."""
+    observable_times: set[float] = set()
+    default_times = config.default_evaluation_times
```
a-quelle-pasqal: the default times should be used instead of `obs.evaluation_times` if the latter does not exist. It should not be used to prefill like this. See the old logic from `_get_target_times`

emu_base/pulser_adapter.py
```diff
@@ -38,30 +39,44 @@ def _get_all_lindblad_noise_operators(
...
+    then converts everything to absolute times over the sequence duration.
+    """
+    duration = float(sequence.get_duration(include_fall_time=config.with_modulation))
+    n_steps = math.floor(duration / dt)
+    evolution_times_rel: set[float] = {
+        i * float(dt) / duration for i in range(n_steps + 1)
+    }
+    target_times_rel = evolution_times_rel | _unique_observable_times(config)
+    target_times: list[float] = sorted({t * duration for t in target_times_rel})
+    assert target_times[-1].is_integer()  # pulser requires int duration
```
a-quelle-pasqal: I don't think you can safely assert on this, since `duration` might be missing from the list. in `evaluation_times_rel` `1.0` will not be present if `dt` does not divide `duration` and in the obsevables, you cannot rely on `1.0` being an evaluation time either. See my comment above.

PR #212, opened 2026-02-03
doc-checker in the pre-commit

murogrande:
integrating doc-checker to the current pre-commit.
There are 2 hooks:
```
    - id: doc-checker-basic
    language: system
    args: ["--warn-only","--ignore-submodules", "emu_mps.optimatrix"]
```

     - id: doc-checker-links
       language: system
       args: ["--ignore-submodules", "emu_mps.optimatrix"]
       verbose: true
    
- the doc-checker-basic: checks if the api in docs contains all the classes, look for internal broken links and parameters in functions (only warning for the moment, which means that this will not stop for committing)
- the doc-checker-link: will search for broken external links

for more information check [doc-checker](https://github.com/murogrande/doc_checker/tree/main)

TODO: 
- LLM backend part where doc-checker analyse the docstrings will be in a second phase.


.pre-commit-config.yaml
```diff
@@ -23,14 +23,25 @@ repos:
...
     hooks:
       - id: mypy
         exclude: examples|test|ci
         additional_dependencies: ["pulser-core[torch]>=1.7.0", "torch>=2.9.0"]  # The version in pyproject.toml must match
+
+  - repo: https://github.com/pasqal-io/doc_checker
+    rev: 53bb9d79b85d2bdc8ec7ffbc05b97749c3593583 # full SHA for the commit to use
+    hooks:
+     - id: doc-checker-basic
+       language: system
```
elmerhej: If I understood this correctly,  `system` tells the pre-commit to run the doc-checker, since it was already installed via pip in the `lint_test.yml`.

but why do we need it? I think we can remove this line here, then `doc-checker` will be installed automatically by ppre-commit in an isolated env, and then we won't need to install it in lint_test.yml.

elmerhej (approved):

elmerhej:
LGTM

PR #223, opened 2026-02-20
Adding restart for non converging Krylov algorithm

kbidzhiev:
Summary

This PR updates the Lanczos implementation to use a residual-minimizing restart strategy for DMRG only. After each iteration the algorithm evaluates the residuals of the candidate vectors, selects the vector with the smallest residual, and restarts the search from that vector.

What changed

- Added residual-based selection across the current set of candidate vectors.
- Modified the restart logic so the next Lanczos run begins from the best (lowest-residual) vector instead of continuing from the last generated vector.
- Updated the iteration flow to support “pick-best-and-restart” without altering the external API/inputs.

exhaustive guide to numerical errors etc in Lanczos algo 
Lectures on Lanczos with detail study of residual
https://www.stat.uchicago.edu/~lekheng/courses/302/demmel/demmch7.pdf


emu_base/math/krylov_energy_min.py
```diff
@@ -4,104 +4,187 @@
...
-                c * vec for c, vec in zip(ground_eigenvector, lanczos_vectors)
-            )
-            final_state = final_state / final_state.norm()
-            happy_breakdown = True
-            converged = True
-            break
 
-        # Reconstruct final state in original Hilbert space
-        lanczos_vectors.append(w / beta)
-        final_state = sum(c * vec for c, vec in zip(ground_eigenvector, lanczos_vectors))
```
kbidzhiev, line 89: size of `ground_eigenvector` is old, but `lanczos_vectors` are updated.
now they have diff size

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,227 @@
...
+    v = torch.tensordot(c, Q, dims=([0], [0]))
 
-def krylov_energy_minimization_impl(
+    norm = v.norm()
+    if norm.item() <= NUMERICAL_TOLERANCE:
+        raise ValueError("Ritz vector has zero norm")
+    return cast(torch.Tensor, v / norm)  # mypy requires explicit type
+
+
+def build_next_lanczos_vector(
```
a-quelle-pasqal: The name of this function is not descriptive of what it actually does, which is building the next lanczos vector AND filling the T matrix

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,227 @@
...
+def _ritz_vector(coefficients: torch.Tensor, basis: list[torch.Tensor]) -> torch.Tensor:
+    """
+    Return the normalized Ritz vector from Lanczos basis vectors.
+    """
+    if len(coefficients) != len(basis):
+        raise ValueError(
+            f"Expected len(coefficients) == len(basis), got {len(coefficients)} != {len(basis)}"
+        )
 
+    Q = torch.stack(basis)
```
a-quelle-pasqal: I dislike this, it duplicates all the basis vectors in memory, which can mean a lot of memory is wasted, since storing the Lanczos vectors is where most of the memory in the algorithm is used. The original implementation with `sum` was better.

emu_base/math/krylov_energy_min.py
a-quelle-pasqal: zip only goes as far as the shortes of the two lists. I think they were always of different size, and that is ok?

a-quelle-pasqal (changes requested):
Please also add the literature, I'm especially interested in the justification for
`resid = (beta * y[-1]).abs()  # == norm(op(ritz_vec) - ritz_value * ritz_vec)`

test/emu_base/math/test_krylov_energy_min.py
```diff
@@ -201,12 +205,120 @@ def test_live_data(lb, rb, combined_hamiltonian_factors, v, krylov_dim):
...
+    expected_e0 = torch.tensor(-1.0)
+    expected_v0 = torch.tensor([1.0, -1.0]) / torch.sqrt(torch.tensor(2))
+    sgn = expected_v0 @ v0
+    expected_v0 *= sgn
+    assert torch.allclose(e0, expected_e0)
+    assert torch.allclose(v0, expected_v0)
+
+
+def test_lowest_eigen_pair_rejects_non_hermitian():
+    h = torch.tensor([[1.0, 2.0], [0.0, 1.0]])
```
a-quelle-pasqal: I'm not really sure we need the asserts in `lowest_eigen_pair` an din `ritz_vector`. We have full control over what goes into those functions, and it shouldn't ever trigger those checks, while they do take time to compute.

emu_base/math/krylov_energy_min.py
kbidzhiev: 🚀

emu_base/math/krylov_energy_min.py
kbidzhiev: They must be the same size, since this is a dot product.
My comment on the old code is meant to highlight our previous misunderstanding of how the next vector is built and then it updates lanczos basis

test/emu_base/math/test_krylov_energy_min.py
kbidzhiev: refactored and removed tests

We previously ran into an issue where T_matrix was silently not Hermitian. I now build tridiag explicitly and kept the comment as a reminder for potential future use with LOBPCG or other methods.

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,227 @@
...
+    return torch.vdot(a.reshape(-1), b.reshape(-1))
+
+
+def _lowest_eigen_pair(h: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
     """
-    Return the lowest eigenpair of the hermitian matrix T_trunc.
+    Return the lowest eigenpair of the hermitian matrix h.
     """
-    eig_energy, eig_state = torch.linalg.eigh(T_trunc)
+    err = (h - h.mH).norm()
```
elmerhej: why don't you `torch.allclose(h, h.T)` instead? since the truncated matrix is built symmetrically anyway

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,227 @@
...
+    ground_state: torch.Tensor
+    ground_energy: float
+    residual_norm: float
+    converged: bool
+    happy_breakdown: bool
+    iteration_count: int
+    restart_count: int
+
+
+def _dotc(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
```
elmerhej: I think this function is redundant

emu_base/math/krylov_energy_min.py
kbidzhiev: fair. done

emu_base/math/krylov_energy_min.py
kbidzhiev: refactored. the `h` matrix is build "on fly"

kbidzhiev:
> Please also add the literature, I'm especially interested in the justification for `resid = (beta * y[-1]).abs() # == norm(op(ritz_vec) - ritz_value * ritz_vec)`

see Saad (1), Prop. 6.8 (Ch. 6, p. 131).

(1) Numerical Methods for Large Eigenvalue Problems, Y. Saad.
https://epubs.siam.org/doi/book/10.1137/1.9781611970739

emu_base/math/krylov_energy_min.py
kbidzhiev: it's just a wrapper, I tried to avoid explicit reshape and conjugation for the `lhs` since `torch.vdot` does it automatically

kbidzhiev: removed

elmerhej (dismissed):

test/emu_base/math/test_krylov_energy_min.py
a-quelle-pasqal: could you check that these tests now converge for a lower max krylov dim? Say 50?

test/emu_base/math/test_krylov_energy_min.py
a-quelle-pasqal: The comment is good, but the asserts are still in the code. Are you arguing for also keeping them?

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,242 @@
...
+    Restart strategy (explicit):
+      - Run a Lanczos cycle up to `max_krylov_dim`.
+      - If not converged, restart with the best Ritz vector (lowest residual) from that cycle.
+
+    Convergence:
+      - residual norm ||Hψ - Eψ|| < residual_tolerance
+      - happy breakdown: beta < norm_tolerance
+    """
+
+    result = KrylovEnergyResult(
```
a-quelle-pasqal: this is not used? There is result returned in line `212`

emu_base/math/krylov_energy_min.py
```diff
@@ -1,121 +1,242 @@
...
+        total_iters += result.iteration_count
+        result = replace(result, restart_count=r, iteration_count=total_iters)
+
+        if result.happy_breakdown or result.converged:
+            break
+
+    return result
+
+
 def krylov_energy_minimization(
```
a-quelle-pasqal: I had a discussion with Elie, and it would make the file easier to read if you reverse the order in which the functions are defined in the file. The entry point here will be `krylov_energy_minimization` so if you put this at the top, it will be where the user starts reading. Then they will see it calls `krylov_energy_minimization_impl` so this can be the function below in the file etc.

test/emu_base/math/test_krylov_energy_min.py
kbidzhiev: removed

test/emu_base/math/test_krylov_energy_min.py
kbidzhiev: done

emu_base/math/krylov_energy_min.py
kbidzhiev: I create a variable outside of the scope, so the loop look this way 

```
res = 0
for _ in range(..):
    new_res = update_res(res)
    res = new_res
```

emu_base/math/krylov_energy_min.py
kbidzhiev: done

kbidzhiev:
> Please also add the literature, I'm especially interested in the justification for `resid = (beta * y[-1]).abs() # == norm(op(ritz_vec) - ritz_value * ritz_vec)`

done in the comments in the code and in the head of the file

emu_base/math/krylov_energy_min.py
elmerhej: Thanks for making the change :+1:

elmerhej:
LGTM

a-quelle-pasqal (approved):
LGTM

PR #228, opened 2026-03-12
Bath compression

### LINKED ISSUE #95: Bath compression
We can try compressing the baths with ~SVD.
Every bath element is a tensor with 3 legs $`(\chi, mH_i, \chi)`$ meaning 3d tensor is a list of Hermitian matrices of length $`mH_i`$

These matrices can be decomposed with `torch.eigh()` $`B_{\chi, \chi} = U S U^\dagger`$ and small eigenvalues(~singular values) can be chopped off, let's say `1e-10`.

Again, this is a memory-performance trade. Maybe Multiplication $`H_eff \psi`$ will happen faster

kbidzhiev:
In this PR, I introduce a new method for packing Hermitian objects of the shape (χ, m , χ), specifically energy baths.

The main idea is to avoid storing the full Hermitian matrix. Instead, we store only the lower triangular part, including the main diagonal. The upper triangular part can then be reconstructed from the lower triangular entries using complex conjugation. Lower triangular columns are stacked to `m` vectors of length $$\chi(\chi+1)/2$$

This reduces the memory footprint of an $$\chi \times \chi$$ matrix from $$\chi^2$$ to $$\chi(\chi+1)/2$$.

The PR closes https://github.com/pasqal-io/emulators/issues/95
(I've tried decompositions like `SVD`, `eigh`, `cholesky`. No compression, baths are ~full rank tensors)

the following part contains results for 5x5 `quench_performance_gpu` from our benchmark suit.
Memory measurement is done with `get_max_rss` in emu_base/utils.py, which is using `torch.cuda.max_memory_allocated(...)`. In the following I'm going to tests the new code with `.memory_allocated(...)` vs `.max_memory_allocated(...)`
### Results

`torch.cuda.memory_allocated(...)`: 
`Returns the current GPU memory occupied by tensors in bytes for a given device.`
<table>
  <tr>
    <td align="center">
      <b>Before</b><br>
      image
    </td>
    <td align="center">
      <b>After</b><br>
      image
    </td>
  </tr>
</table>

`torch.cuda.max_memory_allocated(...)`:  <- MAX allocation:
`Return the maximum GPU memory occupied by tensors in bytes for a given device.
    By default, this returns the peak allocated memory since the beginning of
    this program.`

<table>
  <tr>
    <td align="center">
      <b>Before</b><br>
      image
    </td>
    <td align="center">
      <b>After</b><br>
      image
    </td>
  </tr>
</table>

#### NB on performance. 
every Hermitean bath tensor is validated on hermiticity. That can be weakened

### Outcome

The real GPU memory occupation is reduced asymptotically by factor 2, though the peak memory occupation is still high and is not addressed here. The peak memory is hidden in operator application in Krylov routines


emu_base/__init__.py
```diff
@@ -1,26 +1,28 @@
 from .constants import DEVICE_COUNT
-from .pulser_adapter import PulserData, HamiltonianType, SequenceData
-from .math.brents_root_finding import find_root_brents
-from .math.krylov_exp import krylov_exp, DEFAULT_MAX_KRYLOV_DIM
 from .jump_lindblad_operators import compute_noise_from_lindbladians
+from .math.brents_root_finding import find_root_brents
```
kbidzhiev, line 3: the file is sorted in alphabetic order

a-quelle-pasqal:
I'm closing this, since it's made obsolete by #256 which creates a better improvement at lower runtime cost.

PR #229, opened 2026-03-12
TDVP small refactor

kbidzhiev:
I've simplified logic in TDVP, making a first step toward generalisation.

- `_swipe_direction`, `_sweep_index` and `_timestep_index` renamed because they are internal variables of the class
- `_swipe_direction`, `_sweep_index` and `_timestep_index` have default values and its reflected in a default constructor
- Sweep direction and MPS node indexing are logically not coupled, its now reflected in the code


a-quelle-pasqal (approved):
LGTM

PR #237, opened 2026-05-01
add SequenceData API layer

a-quelle-pasqal:
I slightly refactored the `Backend.run` function to depend on `Backend._run_from_sequence_data(SequenceData, Config)` for both `SVBackend` and `MPSBackend`. I also added a test for both.

This will allow users to define a `SequenceData` directly and run from there (this will be useful for the counter-diabatic driving project). Note that the `SequenceData` contains `target_times` and any `Observable.evaluation_time` that is not in `target_times` will be ignored. When going through `run`, `PulserData` will add `evaluation_times` to `target_times`, but that part of the workflow is now circumvented.

Also, I'm setting an upper bound on the torch version for running on the cluster.


emu_base/pulser_adapter.py
```diff
@@ -189,12 +189,15 @@ class SequenceData:
...
-    noise_model: NoiseModel
-    qubit_ids: tuple[QubitId, ...]
+    state_prep_error: float
     target_times: list[float]
     eigenstates: list[States]
     hamiltonian_type: HamiltonianType
 
+    @property
+    def qubit_ids(self) -> tuple[QubitId, ...]:
+        return tuple(self.bad_atoms.keys())
```
elmerhej: It feels now that you've inverted the roles of both `qubit_ids` and `bad_atoms`. Now we require `bad_atoms` to define `qubit_ids`.

and since `SequenceData` is intended to become part of the public API, the user will need to define the qubit ids from the `bad_atoms` field, which isn't very explicit.

besides, this
`SequenceData(qubit_ids=("q0","q1"), ...)` is much much more intuitive to the user than below

`SequenceData(bad_atoms={"q0": False, "q1": False}, ...)`

elmerhej: would it be better to keep `qubit_ids` as an explicit field?

elmerhej:
One possible alternative design could be to accept an optional SequenceData argument in `run()`,  where either `Pulser` is used or a user-constructed `SequenceData` is provided directly. In this way, we can have a single public entry point (`run`).

PR #238, opened 2026-05-04
added dmm noise types

a-quelle-pasqal:
The new dmm noise types added in the noise model need to be added in emu-base so that it doesn't think they're lindblad noises. Required to make the following code work

To see what this supports look at the modified test.


elmerhej (approved):
LGTM

PR #240, opened 2026-05-11
change torch lower version

a-quelle-pasqal:
The ops team couldn't get torch >2.7 to work on one of the machines that we really have to support. I lowered the upper bound of torch to include 2.7. Initially I wanted to use the same bound as in pulser core (2.6) but some of the tests started failing.

I also removed the upper bound for the released packages, since there were some complaints. I kept the bound for installing from source, since it's needed for the benchmarks.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
-version: 2.7.4
+version: 2.7.5
 date-released: '2025-05-12'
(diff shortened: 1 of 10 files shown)


CITATION.cff
```diff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
-version: 2.7.4
+version: 2.7.5
 date-released: '2025-05-12'
```
elmerhej: should we update the release date? if someone cites our repo, it'll show 2025 as the year, which is wrong

elmerhej (approved):
LGTM

PR #241, opened 2026-05-19
Clean up old right baths before computing new ones

a-quelle-pasqal:
When starting a new TDVP sweep, emu-mps was temprarily keeping both the old and the new right baths in memory. Usually, this is not a problem since the Lanczos algorithm is the most memory-intensive part of the program, but for larger qubit numbers, there are situations where this causes the program to OOM when creating the new baths.

This MR fixes the issue.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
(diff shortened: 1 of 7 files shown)


elmerhej (approved):
LGTM

PR #243, opened 2026-05-28
Reorder single qubit driving terms in emu-mps

a-quelle-pasqal:
When `MPSConfig.optimize_qubit_ordering == True`, the single qubit driving terms in the Hamiltonian need to be reordered along with the interaction matrix, the initial state etc. In `MPSBackendImpl` this is now done.

Independently, there was an issue in `PulserAdapter` where qubits were missing in the `omega`, `delta` and `phi` tensors. This caused the driving terms to added to the incorrect qubit in the Hamiltonian regardless of qubit reordering. I also fixed this issue, and there are tests.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
-version: 2.7.6
+version: 2.7.7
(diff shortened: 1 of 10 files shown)


elmerhej (dismissed):
LGTM

PR #245, opened 2026-06-01
As/torch device mismatch

alessandro-santini:
## Small fix for GPU device support for TDVP

Both krylov_exp_impl and MPS.expect_batch build an intermediate tensor with torch.zeros(...) but don't pass a device, so it always lands on CPU. On GPU that means these tensors sit on a different device than the data they're combined with, which should slow down the computation.

- krylov_exp.py: T now takes device=v.device
- mps.py: result now takes device=self.factors[0].device

The mps.py one should be especially handy for noisy simulations, where you typically run many trajectories on GPU and call expect_batch for the Kraus jump operators.

For reference, the lanczos helper in double_krylov.py already allocates T on the right device (device=v.device), so it's consistent with this change.

I tested it on my Mac's GPU (Metal), which partially confirms the device fix works, but since Metal isn't a full CUDA substitute a check on performance on a real CUDA device would be good.

### DIFF
--- a/emu_base/math/krylov_exp.py
+++ b/emu_base/math/krylov_exp.py
@@ -44,7 +44,9 @@ def krylov_exp_impl(
     v /= initial_norm
 
     lanczos_vectors = [v]
-    T = torch.zeros(max_krylov_dim + 2, max_krylov_dim + 2, dtype=v.dtype)
+    T = torch.zeros(
+        max_krylov_dim + 2, max_krylov_dim + 2, dtype=v.dtype, device=v.device
+    )
 
     for j in range(max_krylov_dim):
         w = op(lanczos_vectors[-1])
--- a/emu_mps/mps.py
+++ b/emu_mps/mps.py
@@ -507,7 +507,12 @@ class MPS(State[complex, torch.Tensor]):
             else self.orthogonalize(0)
         )
 
-        result = torch.zeros(self.num_sites, single_qubit_operators.shape[0], dtype=dtype)
+        result = torch.zeros(
+            self.num_sites,
+            single_qubit_operators.shape[0],
+            dtype=dtype,
+            device=self.factors[0].device,
+        )
 
         center_factor = self.factors[orthogonality_center]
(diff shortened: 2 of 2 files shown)


emu_mps/mps.py
```diff
@@ -507,7 +507,12 @@ def expect_batch(self, single_qubit_operators: torch.Tensor) -> torch.Tensor:
             else self.orthogonalize(0)
         )
 
-        result = torch.zeros(self.num_sites, single_qubit_operators.shape[0], dtype=dtype)
+        result = torch.zeros(
```
a-quelle-pasqal: we always want result to be output to the cpu for compatibility reasons. I don't think there is any loss in creating this tensor on cpu, since the data will have to be moved there anyway.

emu_base/math/krylov_exp.py
```diff
@@ -44,7 +44,9 @@ def krylov_exp_impl(
     v /= initial_norm
 
     lanczos_vectors = [v]
-    T = torch.zeros(max_krylov_dim + 2, max_krylov_dim + 2, dtype=v.dtype)
+    T = torch.zeros(
```
a-quelle-pasqal, line 47: This could be a valid improvement. We have a benchmark suite that we run on the cluster, but you don't have access to it yet since it's in gitlab. I will run the benchmarks. If it shows a clear improvement, we can merge your code.

a-quelle-pasqal: I've run the benchmarks on our cluster and it's actually slower. The likely reason is that we have to compute the matrix exponential of `T`, and the A100 gpus on the cluster have a much higher degree of paralellism that gpu in your mac, but also a much lower clock speed. This makes the A100s much faster on very large data, but much slower on small data, and the `T` matrix is typically about 30x30, which is quite small.

To keep the scope of the library manageable, we're really optimizing for predominantly for our cluster, which is already difficult enough. If you agree with the above, please close the PR. Thank you for your effort though, we do still find performance improvements in our code occasionally, so it's worth taking seriously.

PR #247, opened 2026-06-17
cast numpy.bool to bool in emu-base

a-quelle-pasqal:
older versions of torch dont handle `numpy.bool` correctly. Bad atoms are currently specified as a `numpy.bool`. In this MR, I cast them to normal `bool` in `PulserAdapter` so we don't have to worry about the types in the emulators.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -40,4 +40,4 @@ keywords:
   - analog quantum computing
   - emulation
 version: 2.7.8
-date-released: '2026-05-11'
+date-released: '2026-06-18'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
-version: 2.7.7
+version: 2.7.8
 date-released: '2026-05-11'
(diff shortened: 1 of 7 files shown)


emu_base/pulser_adapter.py
```diff
@@ -317,7 +317,10 @@ def get_sequences(self) -> Iterator[SequenceData]:
                     phi,
                     interaction_matrix,
                     tuple(samples.trajectory.bad_atoms.keys()),
-                    tuple(samples.trajectory.bad_atoms.values()),
+                    tuple(
```
murogrande: probably, minor comment
tuple(bool(x) for x in samples.bad_atoms.values())

murogrande (commented):
small improvement

CITATION.cff
```diff
@@ -39,5 +39,5 @@ abstract: >-
 keywords:
   - analog quantum computing
   - emulation
-version: 2.7.7
+version: 2.7.8
 date-released: '2026-05-11'
```
murogrande: update this?

elmerhej: Since we often forget about the `release-date`, could it somehow be automated?

elmerhej (approved):
LGTM

PR #248, opened 2026-06-23
Generalize Hamiltonian MPO logic in emu-mps

a-quelle-pasqal:
The API now accepts a stack of interaction matrices, and the corresponding 1-body operator so it can add several terms of the form $\sum_{i,j} U_{ij}O_i O_j$ for single body operator $O$. This is the most significant change, which happens in `hamiltonian.py`. Furthermore, in anticipation of pulser 1.9, pulser_adapter now returns interaction matrices as  rank 3 tensors, being a stack of itneraction matrices. All the other changes just propagate the corresponding API changes through the rest of the code.

In pulser 1.9, a Rydberg term will be added for XY mode, and the interaction matrix will be a stack as above. The other needed modifications will become trivial after this.


emu_mps/hamiltonian.py
```diff
@@ -63,343 +68,259 @@ def __iter__(self) -> Iterator[torch.Tensor]:
...
                 i += 1
-            j += 1
+            for (
+                current_right_interaction
+            ) in current_right_interactions_per_int.nonzero().flatten():
+                if right_interactions_to_keep[ind][current_right_interaction]:
+                    factor[i, :, :, j] = self.identity
+                    i += 1
+                else:
+                    print("not")
```
murogrande: print

a-quelle-pasqal: Yes, the entire else should be removed, it's there for debugging, and I forgot to take it out after.

emu_mps/hamiltonian.py
```diff
@@ -21,31 +19,38 @@ class Operators:
     sy = torch.tensor([[0.0, -0.5j], [0.5j, 0.0]], dtype=dtype)
 
 
-class HamiltonianMPOFactors(ABC):
+class HamiltonianMPOFactors:
     """Abstract class for MPO factors of a two-body Hamiltonian.
```
elmerhej: this is no longer an abstract class. should we update the docstring as a result?

emu_mps/hamiltonian.py
```diff
@@ -21,31 +19,38 @@ class Operators:
...
+        dim: int = 2,
+    ):
         self._validate_interaction_matrix(interaction_matrix)
 
         if dim not in (2, 3):
             raise ValueError(f"dim must be 2 or 3, got {dim}")
         self.dim = dim
 
         self.interaction_matrix = interaction_matrix.clone()
-        self.interaction_matrix.fill_diagonal_(0.0)  # or assert
```
elmerhej, line 39: that's dropped now. I presume pulser will be producing zero diagonal matrices (i.e. `U_i,i = 0` for all U), and if that's the case, shouldn't we at least add an assertion in e.g. `_validate_interaction_matrix`?

emu_mps/hamiltonian.py
elmerhej, line 325: should we also here tweak the docstrings?

emu_mps/mps_backend_impl.py
```diff
@@ -138,6 +140,23 @@ def __init__(self, mps_config: MPSConfig, pulser_data: SequenceData):
...
         self.timestep_count: int = self.omega.shape[0]
 
+        # move to emu-base when emu-sv supports non-rydberg?
+        interaction_ops: torch.Tensor
+        if pulser_data.hamiltonian_type == HamiltonianType.Rydberg:
+            interaction_ops = torch.tensor(
+                [[[0.0, 0.0], [0.0, 1.0]]], dtype=torch.float64
+            )
+        elif pulser_data.hamiltonian_type == HamiltonianType.XY:
+            interaction_ops = np.sqrt(2) * torch.tensor(
```
elmerhej: you are using numpy only once in the entire file. That's unnecessary. we can replace it with `2**0.5` and avoid importing `numpy`

emu_mps/mps_backend_impl.py
```diff
@@ -138,6 +140,23 @@ def __init__(self, mps_config: MPSConfig, pulser_data: SequenceData):
         self.phi = pulser_data.phi[:, self.qubit_permutation]
         self.timestep_count: int = self.omega.shape[0]
 
+        # move to emu-base when emu-sv supports non-rydberg?
```
elmerhej, line 142: we should remember this. how about it we add it as an issue?

emu_sv/sv_backend_impl.py
```diff
@@ -139,8 +139,8 @@ def init_dark_qubits(self) -> None:
 
             def interaction_matrix(t: float) -> torch.Tensor:
                 mat = original(t).clone()
-                mat[indices, :] = 0.0
-                mat[:, indices] = 0.0
+                mat[:, indices, :] = 0.0
```
elmerhej: maybe `mat[:, indices] = 0.0`?

emu_mps/mps_backend_impl.py
murogrande: the same as Elie
minor comment: probably 2**0.5 or math.sqrt(2) and drop the import ?

emu_mps/hamiltonian.py
```diff
@@ -21,31 +19,38 @@ class Operators:
...
     """Abstract class for MPO factors of a two-body Hamiltonian.
 
     Subclasses implement the local MPO tensor at each position in the chain:
     first site, left half, middle, right half, and last site.
     """
 
-    def __init__(self, interaction_matrix: torch.Tensor, dim: int = 2):
+    def __init__(
+        self,
+        interaction_matrix: torch.Tensor,
```
murogrande, line 30: Did you try the code feeding an `interaction_matrix` in GPU?

emu_mps/hamiltonian.py
a-quelle-pasqal: I will think a bit about what is best, but for code safety reasons I agree we should do something.

emu_mps/mps_backend_impl.py
a-quelle-pasqal: I created #249

emu_sv/sv_backend_impl.py
a-quelle-pasqal: I'll see if that works, I guess it should.

emu_mps/mps_backend_impl.py
elmerhej: super cool, thanks.
I will resolve

emu_mps/hamiltonian.py
murogrande: this argument is not valid any more (and doc-checker did not mentioned this when I run it, probably because this is not a public function)

emu_mps/hamiltonian.py
```diff
@@ -63,343 +68,257 @@ def __iter__(self) -> Iterator[torch.Tensor]:
...
-        right_bond_dim = 4 if has_right_interaction else 2
-        factor = self._empty_factor(left_bond_dim, right_bond_dim)
-        factor[0, :, :, 1] = self.identity
-        if has_right_interaction:
-            factor[0, :2, :2, 2] = Operators.sx
-            factor[0, :2, :2, 3] = Operators.sy
-
-        return factor
+    def _has_left_interaction(self, site: int) -> list[bool]:
+        """Returns a Tensor filled with bool, one for each interaction term"""
```
murogrande: Returns a list of bool

emu_mps/mps_backend_impl.py
```diff
@@ -138,6 +140,23 @@ def __init__(self, mps_config: MPSConfig, pulser_data: SequenceData):
...
+        interaction_ops: torch.Tensor
+        if pulser_data.hamiltonian_type == HamiltonianType.Rydberg:
+            interaction_ops = torch.tensor(
+                [[[0.0, 0.0], [0.0, 1.0]]], dtype=torch.float64
+            )
+        elif pulser_data.hamiltonian_type == HamiltonianType.XY:
+            interaction_ops = np.sqrt(2) * torch.tensor(
+                [[[0.0, 0.5], [0.5, 0.0]], [[0.0, -0.5j], [0.5j, 0.0]]],
+                dtype=torch.complex128,
```
murogrande, line 151: XY dtype should be the same as the Rydberg, torch.complex64

emu_mps/hamiltonian.py
a-quelle-pasqal: The diagonals of the interaction matrix are not used. Since the values do not impact the code, I think we can skip the check altogether without problem.

elmerhej: you're right, I agree. 
thanks :+1:

emu_mps/mps_backend_impl.py
elmerhej: why? I think we should stick to `torch.complex128` for both

a-quelle-pasqal: The XY term needs to be `complex128`, it can't be `float64` because the $$S_y$$ operator is not real. To make them the same I changed the dtype of the Rydberg term `float64 -> compex128`

emu_mps/hamiltonian.py
```diff
@@ -37,7 +37,7 @@ def __init__(
             raise ValueError(f"dim must be 2 or 3, got {dim}")
         self.dim = dim
 
-        self.interaction_matrix = interaction_matrix.clone()
+        self.interaction_matrix = interaction_matrix.clone().cpu()
```
elmerhej, line 40: isn't the interaction matrix already built on the CPU? so `.cpu()` essentially does nothing

a-quelle-pasqal: it's possible that the user specifies it in the config, and then anything goes

elmerhej: got it, good to know :+1: 
Thanks

emu_mps/hamiltonian.py
murogrande: There is a test related to this comment

